PR #17, opened 2018-04-23
Improve README

dennisklein:
This is not finished yet, but would like some feedback on the paragraph I added so far. Thx


README.md
```diff
@@ -1,6 +1,22 @@
...
 
-C++ Message Queuing Library
+C++ Message Queuing Library and Framework
+
+FairMQ was designed to help creating large-scale processing pipelines needed in next-generation High Energy
+Physics Experiments. FairMQ is written in C++ and aims 
+  * to be **easy to use** (hide the details of low level data transport technologies),
+  * to provide a reasonably **efficient data transport** (zero-copy, high throughput) service,
+  * that is **data format agnostic**, and
+  * to provide some of the **basic building blocks** to implement a higher level data processing framework on top.
```
rbx: Even though you describe this in the following paragraph, I would add also here one more point about the abstract interface that allows implementation with efficient current and future transport technologies.

README.md
```diff
@@ -1,6 +1,22 @@
...
+
+The core of FairMQ provides an abstract message queuing API with scalability protocols inspired by [ZeroMQ](https://github.com/zeromq/libzmq) (e.g. PUSH/PULL, PUB/SUB). 
+FairMQ provides multiple implementations for its API (so-called "transports", e.g. `zeromq`, `shmem`, `nanomsg`, and `ofi` (in development)) to cover a variaty of use cases 
+(e.g. inter-thread, inter-process, inter-node communication) and machines (e.g. Ethernet, Infiniband). 
+In addition to this core functionality FairMQ provides a framework for creating "devices" - actors which 
+are communicating through message queuing. Each device is equipped with 
+a high level state machine, that enables the customization and specialization with user tasks, and a plugin system 
+for configuration and control. Next to the provided plugins (e.g. [DDS](https://github.com/FairRootGroup/DDS)) the user can develop her own plugins 
+to integrate her devices with external configuration and control services.
```
rbx: Here I would add one more sentence that several generic devices are already provided by us.

Issue #13, opened 2018-04-18
Move example code from FairRoot repo to this repo


dennisklein:

dennisklein:
Done.

PR #18, opened 2018-04-27
Move MQ examples from FairRoot to this repository

rbx:


rbx:
Resolves #13 and #19.

Issue #4, opened 2018-04-13
Implement config plugin


dennisklein:
- [x] Move existing config logic into a builtin plugin.
- [x] Initialize FairMQProgOptions in the device constructor.
- [x] Support `RemoveProperty()` or similar.
- [x] Create channels implicitely if some attribute (e.g. `chans.asdf.0.address`) is set
- [x] Support all channel properties with `--channel-config` (again?). E.g. `rateLogging=` does not work.

PR #30, opened 2018-05-07
Cleanup of transport management in device/channel

rbx:
- Use directly available `fTransportFactory` as the return value of `FairMQDevice::Transport()` instead of accessing the map.
- Fix namespaces in Transport.h
- Add conversion maps for string/enum version of transport labels.
- In Channel/Device the enum transport labels can be used instead of string.

### DIFF
--- a/fairmq/FairMQDevice.h
+++ b/fairmq/FairMQDevice.h
@@ -253,7 +253,7 @@ class FairMQDevice : public FairMQStateMachine
         // if more than one channel provided, check compatibility
         if (chans.size() > 1)
         {
-            FairMQ::Transport type = fChannels.at(chans.at(0)).at(0).Transport()->GetType();
+            fair::mq::Transport type = fChannels.at(chans.at(0)).at(0).Transport()->GetType();
 
             for (unsigned int i = 1; i < chans.size(); ++i)
             {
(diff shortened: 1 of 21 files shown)


fairmq/FairMQDevice.h
```diff
@@ -196,7 +196,7 @@ class FairMQDevice : public FairMQStateMachine
     /// @brief Getter for default transport factory
     auto Transport() const -> const FairMQTransportFactory*
     {
-        return fTransports.at(fair::mq::TransportTypes[GetDefaultTransport()]).get();
+        return fTransportFactory.get();;
```
dennisklein: Remove one semicolon

fairmq/Transports.h
```diff
@@ -37,25 +53,15 @@ static std::unordered_map<std::string, Transport> TransportTypes {
...
-
-namespace fair
-{
-namespace mq
-{
-
-using Transport = ::FairMQ::Transport;
-using ::FairMQ::TransportTypes;
+static std::unordered_map<Transport, std::string> TransportNames {
+    { Transport::DEFAULT, "default" },
```
dennisklein: Why do we need `Transport::DEFAULT` again?

dennisklein: As discussed, let's keep it like this and remove the `Transport::DEFAULT` entry, once we resolve #4.

fairmq/FairMQDevice.h
rbx: Fixed.

PR #45, opened 2018-06-08
Avoid copy (where possible) when switching transports

rbx:
- Avoid copy when transport message buffer can be directly taken over by the target transport. E.g. shared memory message can be given to zmq without copy. This is done by using the buffer+size+callback message creator, for which each transport already knows if a copy is needed or not.
- improve some error handling/output in socket/device/state machine.

---

Checklist:


dennisklein:
Somehow your merge commit did not trigger alfaci, triggered it manually now.

PR #49, opened 2018-06-12
Make UpdateTransport log to INFO and not WARN

dberzano:
At the moment the AliceO2 DPL has problems in marking problematic devices
because of the log level too high of such informative messages.

### DIFF
--- a/fairmq/FairMQChannel.cxx
+++ b/fairmq/FairMQChannel.cxx
@@ -327,9 +327,9 @@ void FairMQChannel::UpdateTransport(const string& transport)
     {
         unique_lock<mutex> lock(fChannelMutex);
         fIsValid = false;
-        LOG(WARN) << fName << ": " << transport;
+        LOG(INFO) << fName << ": " << transport;
         fTransportType = fair::mq::TransportTypes.at(transport);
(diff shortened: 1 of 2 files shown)


dberzano:
@dennisklein note that 811e716731ecac1f484fa1c3b3ca5dedbabbe666 is actually tagged as v1.2.4 but it's not on `master`.

dennisklein:
Please open the PR against `dev` as stated in the PR template. Thx!

> @dennisklein note that 811e716 is actually tagged as v1.2.4 but it's not on master.

This is intended. In this repo we have development releases, which are not considered stable (not tested exhaustively etc, but can be used instead of tracking the `dev` branch). At arbitrary points in time we make the effort to stabilize the dev branch, then we merge into `master` and declare a certain tag stable.

dennisklein:
We just noticed your change is already in `dev`, I will add a new development release.

dberzano:
Woops sorry. FTR it's not even needed as the change was already done by yourself in `dev`. Would it be possible to have a tag?

dberzano:
(OK we keep crossing our comments!)

dennisklein:
I added [v1.2.5](https://github.com/FairRootGroup/FairMQ/releases).

dberzano:
Thanks, we're bumping in alisw/alidist#1192 for reference.

PR #55, opened 2018-06-18
Allow ";" as separator in multi-point channel config

mkrzewic:
This is to allow multi-point configuration using Suboptparser - comma is parsed by the option parser and ";" seems a safe alternative.


rbx:
Can we do ";" only perhaps? To avoid ambiguity.

mkrzewic:
why not, done

PR #60, opened 2018-07-19
Add codecov reports to PRs

dennisklein:
Checklist:


dennisklein:
The coverage report is not correct for this PR, but it will be for the following PRs. I had to manually submit a base report so codecov could start comparing against something.

I also configured the bot to update its first comment, if the PR is updated, so nobody gets spammed with notification emails.

Right now, there are no conditions set. But we could configure e.g. that a PR only passes, if the test coverage does not decrease, or that newly added code must be covered to at least x%.

Issue #57, opened 2018-07-11
Turn device algorithm on master thread


sawenzel:
The current choice of letting device algorithm spawn on a non-standard thread (thread 2) is problematic. 

I've had, for instance, problems with certain (thread-aware) classes which I initialize prior to setting up the device and which I then give to the device constructor. Letting the main algorithm turn on the master thread would avoid such problems.

sawenzel:
This is probably related to #46.

Issue #46, opened 2018-06-08
Remove worker thread from state machine


dennisklein:

dennisklein:
will fix #57

PR #66, opened 2018-08-03
Run state handlers on the main thread [breaking change]

rbx:
- Run state handlers (e.g. InitTask(), Run(), ...) on the main thread. This solves some cases that require user code to be run on the main thread (@sawenzel). All the state handlers will be called in the same thread where the newly introduced `device.RunStateMachine()` method is called. It returns when state machine is done (after `END` event). To keep the device responsive when user code is run, state machine controller has to be run in an extra thread. This is a **breaking change** for anyone who does not use our control plugins and has a custom main function. The solution is to add a separate thread either for all the `ChangeState(...)` calls, or for `device.RunStateMachine()`. For an example fix you can see `test/device/_multiple_devices.cxx` inside this PR.
- Reduce number of action handlers in the state machine - reuse same where possible.
- Fix some race conditions inside control plugin.
- Fix control plugin signal handler to comply with signal handler requirements.
- Simplify ownership structure inside DeviceRunner/PluginManager.

---

Checklist:

### DIFF
--- a/test/plugins/_plugin_manager.cxx
+++ b/test/plugins/_plugin_manager.cxx
@@ -15,6 +15,7 @@
 #include <fstream>
 #include <memory>
 #include <vector>
+#include <thread>
 
 namespace
 {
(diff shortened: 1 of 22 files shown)


rbx:
Solves #57, #46.

test/plugins/_plugin_manager.cxx
```diff
@@ -82,7 +87,7 @@ TEST(PluginManager, LoadPluginStatic)
 
     // check order
     const auto expected = vector<string>{"control"};
-    auto actual = vector<string>{};
+    auto actual = vector<string>();
```
dennisklein: You missed the `auto` here ;)

PR #72, opened 2018-08-24
Refactor initialisation & other improvements

rbx:
- Add device constructor that accepts FairMQProgOptions object.
 - Initialize config values in INIT state (to allow their update).
 - Simplify FairMQProgOptions handling in FairMQDevice.
 - Simplify SetTransport/SetConfig - refactor duplicated code.
 - Add FairMQDevice method to add channels.
 - Add RateLimiter class to tools (thanks @mattkretz).
 - Use new RateLimiter in BenchmarkSampler.
 - Use new RateLimiter in ConditionalRun.
 - Add more tests for custom main() - that use SetConfig or SetTransport.

---

Checklist:


fairmq/FairMQDevice.cxx
```diff
@@ -96,16 +145,43 @@ FairMQDevice::FairMQDevice(const fair::mq::tools::Version version)
```
dennisklein: Can't we use delegating constructor here? Furthermore, owning and not owning the config member seems complicated, wouldn't this be a good case for `std::shared_pointer`?

```cpp
FairMQDevice::FairMQDevice()
    : FairMQDevice(std::make_shared<FairMQProgOptions>(), {0, 0, 0})
{
}

FairMQDevice::FairMQDevice(std::shared_ptr<FairMQProgOptions> config)
    : FairMQDevice(config, {0, 0, 0})
{
}

FairMQDevice::FairMQDevice(const fair::mq::tools::Version version)
    : FairMQDevice(std::make_shared<FairMQProgOptions>(), version)
{
}

FairMQDevice::FairMQDevice(std::shared_ptr<FairMQProgOptions> config, const fair::mq::tools::Version version)
    : fTransportFactory(nullptr)
    , fTransports()
    , fChannels()
    , fConfig(config)
    , fId()
    , fInitialValidationFinished(false)
    , fInitialValidationCondition()
    , fInitialValidationMutex()
    , fPortRangeMin(22000)
    , fPortRangeMax(32000)
    , fDefaultTransportType(fair::mq::Transport::DEFAULT)
    , fDataCallbacks(false)
    , fMsgInputs()
    , fMultipartInputs()
    , fMultitransportInputs()
    , fChannelRegistry()
    , fInputChannelKeys()
    , fMultitransportMutex()
    , fMultitransportProceed(false)
    , fVersion(version)
    , fRate(0.)
    , fLastTime(0)
    , fRawCmdLineArgs()
{
}
```

 What do you think?

fairmq/devices/FairMQBenchmarkSampler.cxx
```diff
@@ -1,8 +1,8 @@
 /********************************************************************************
  *    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
```
dennisklein: Let's also update the year, e.g.

` * Copyright (C) 2014-2018 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH *`

dennisklein:
Could you change the order of second and third commit? Thx!

fairmq/FairMQDevice.cxx
rbx: Delegating - yes, will change!

shared_ptr would be a API change, since any external config object would then have to be a shared_ptr, which are now plain objects in all cases. From my perspective this makes user write more code. With two pointers, as proposed (albeit a bit complex internally as you point out), the API stays simple and without break.

fairmq/devices/FairMQBenchmarkSampler.cxx
rbx: ok, will do.

dennisklein:
Merging, codecov.io is experiencing service outages at the moment. so results might not be reliable at the moment anyways.

Issue #82, opened 2018-09-10
sleeping in ConditionalRun inhibits SIGINT


ktf:
I've a device which overrides `ConditionalRun()` and sleeps for a long time. Looks like any external SIGINT / SIGTERM sent to it is delayed until the end of the sleep. Is that expected? How can I have some interruptible sleep for a given device?

rbx:
We catch the signals and let the device finish gracefully. Any Send/Receive calls will return with `-2` and device state will be changed to something other than RUNNING. Indeed your sleep will continue uninterrupted in this case. 

What we could do is to offer something like  `FairMQDevice::WaitFor(chrono::duration)` method that internally will wait on interruption condition variable for the given duration. So for you as a user it will behave just like sleep, but it will be still interruptible by signals and state changes.

Would this help you?

dennisklein:
The new API

```cpp
template<class Rep, class Period>
bool FairMQDevice::WaitFor(std::chrono::duration<Rep, Period> const& duration);
```

is now available for testing in FairMQ `v1.3.3`.

dennisklein:
Thx!

PR #87, opened 2018-09-13
Wait for

rbx:
- Add interruptable FairMQDevice::WaitFor(duration) method (solves #82).
- Add test for FairMQDevice::WaitFor().
- Fix theoretical race in signal handler.
- Use same runner.cxx/.h for all tests.
- Relocate some tests to more apropriate test suits:
   - Poller from Protocols to Poller.
   - Transfer timeout from Protocols to Transport.

---

Checklist:


rbx:
Looks like the Example-Region-nanomsg test became unstable with the new nanomsg version (sometimes it fails).
nanomsg doesn't implement NN_LINGER, thus device doesn't wait at all if there are still outstanding messages in the queue.
I suspect we have never seen a problem with this because for a single message in the tests the shutdown procedure took longer than the sending. Perhaps this has changed with the new version?

rbx:
The Example-Region-nanomsg test is unstable. The sampler quits before sink gets the message. Since the test condition is wether an acknowledgement from the region has been received and in case of nanomsg transport that is a local operation it doesn't have to wait for the sink.
Nanomsg doesn't implement LINGER, thus doesn't wait at all for outstanding transfers before shutting down. In the protocol test suite we have manual sleeps to counteract this, but not in example tests.

Not 100% sure why the problem doesn't appear in other examples, e.g. the 1-1 example. The Example-Region-nanomsg is the only one that gives user buffer directly to nanomsg, perhaps this affects the timing somehow.

fairmq/FairMQDevice.h
```diff
@@ -433,6 +434,16 @@ class FairMQDevice : public FairMQStateMachine
...
     void RunStateMachine() { ProcessWork(); };
 
+    /// Wait for the supplied amount of time or for interruption.
+    /// If interrupted, returns false, otherwise true.
+    /// @param duration wait duration
+    template<class Rep, class Period>
+    bool WaitFor(std::chrono::duration<Rep, Period> const& duration)
+    {
+        std::unique_lock<std::mutex> lock(fInterruptedMtx);
+        return !fInterruptedCV.wait_for(lock, duration, [&] { return fInterrupted.load(); }); // return true if no interruption happened
```
dennisklein: I think you can miss a notification here, e.g. if you wake up spuriously and enter the lambda, read `fInterrupted`, then sleep. Second thread sets `fInterrupted` and wakes you up, but are still in the lambda. You exit the lambda and go to sleep, having consumed the notification at the wrong time. Or?

test/protocols/_waitfor.cxx
```diff
@@ -0,0 +1,74 @@
...
+        cmd << runTestDevice << " --id waitfor_" << " --control static " << " --severity nolog";
+
+        boost::process::ipstream stdout;
+        boost::process::child c(cmd.str(), boost::process::std_out > stdout);
+        string line;
+        getline(stdout, line);
+
+        {
+            std::lock_guard<std::mutex> lock(mtx);
+            pid = std::stoi(line);
```
dennisklein: I think you can get the child process id with `c.id()` here.

test/protocols/_waitfor.cxx
```diff
@@ -0,0 +1,74 @@
...
+ *                  copied verbatim in the file "LICENSE"                       *
+ ********************************************************************************/
+
+#include "runner.h"
+
+#include <gtest/gtest.h>
+#include <boost/process.hpp>
+
+#include <sys/types.h>
+#include <signal.h>
```
dennisklein: I guess those are not used?

fairmq/FairMQDevice.h
rbx: before evaluating the lambda, wait will reacquire the lock, and release it after evaluation. So other threads cannot change fInterrupted (if they lock properly).

test/protocols/_waitfor.cxx
rbx: seems that i missed the id method :) will try tomorrow

test/protocols/_waitfor.cxx
rbx: those are needed for the kill.

test/protocols/_waitfor.cxx
rbx: hmm, but then i will need another sleep. if i just take id from boost, and kill it right away, i will likely kill it before waitfor starts. although there is still some chance of bad timing now (string evaluated and converted before second process proceeds), but much less likely

fairmq/FairMQDevice.h
rbx: it's ⚛️ic. :)

test/protocols/_waitfor.cxx
rbx: Perhaps I am missing it, like with the `id()`, but I only see `terminate()` which sends SIGKILL. I want to have a SIGINT or SIGTERM to give the process a chance to handle it.

fairmq/FairMQDevice.h
dennisklein: > Even if the shared variable is atomic, it must be modified under the mutex in order to correctly publish the modification to the waiting thread.

https://en.cppreference.com/w/cpp/thread/condition_variable

rbx: You're right, missed that point. Adding lock on the writing side.

rbx: Fixed.

test/protocols/_waitfor.cxx
rbx: Changed to using `id()` and removed pid stuff from device.

rbx:
The nanomsg example fails much more now, because with these commits I use `WaitFor()` in the logger thread instead of regular sleep. Regular sleep was giving nanomsg a full second to complete, but now logger exits much sooner.

I'll add a fix the nanomsg test.

rbx:
How about adding a cmd option `--nanomsg-linger` that, if provided, will give nanomsg transport a sleep time before exiting?

dennisklein:
> How about adding a cmd option --nanomsg-linger that, if provided, will give nanomsg transport a sleep time before exiting?

Wouldn't like the name. Can we implement the linger channel option in such a way for nanomsg?

dennisklein:
Flightmode 2h...

rbx:
I think transport option is better than channel one, since if more than one channel has it, it will add the sleeps together. Only one per transport is nedded.

Regarding the name it is in line with other transport specific options, such as `--shm-segment-size`.

rbx:
Ok, it cannot be only in the transport because the sockets are closed already in Reset() before the transport is destroyed. It has to be done in the Socket class, but involving a static member check to avoid multiple sleeps.

rbx:
Ok, I added it as cmd to test, if you have better suggestion how to add it we can change.
I would like to avoid adding transport specific option keys in channel config, right not they are global only. But perhaps we can add `linger` as a general option. Afteral it already exists for ZeroMQ/shmem, but as a hardcoded value. Then I'm fine with adding it to Channel/JSON/SuboptParser.

dennisklein:
Yes, that was my thinking. Have it as a generic option and use it only in nanomsg tests. It can always be a noop implementation for some transports. Having `--nanomsg*` type cmd line options breaks the abstraction.

dennisklein:
Wasn't sure, if linger was configurable yet on channel level. But for a "fake" linger, e.g. for nanomsg, we would just `sleep(max(nn_channels.linger))`. What do you think?

rbx:
Ok, I added linger as a channel option with default value of 500ms. 
For nanomsg internally it is a static member, when nanomsg socket sets linger, it will do std::max(memberLinger, parameterLinger). Once nanomsg socket has lingered, it will remember that it did and will not linger again, unless devices goes after Reset() to Run() again, then it is free to linger again.

fairmq/nanomsg/FairMQTransportFactoryNN.cxx
```diff
@@ -14,10 +14,19 @@ using namespace std;
...
 
-FairMQTransportFactoryNN::FairMQTransportFactoryNN(const string& id, const FairMQProgOptions* /*config*/)
+FairMQTransportFactoryNN::FairMQTransportFactoryNN(const string& id, const FairMQProgOptions* config)
     : FairMQTransportFactory(id)
 {
     LOG(debug) << "Transport: Using nanomsg library";
+
+    if (config)
+    {
+        FairMQSocketNN::SetLinger(config->GetValue<size_t>("nanomsg-linger"));
```
dennisklein: I guess you meant

```cpp
        FairMQSocketNN::SetLinger(config->GetValue<size_t>("linger"));
```

fairmq/nanomsg/FairMQSocketNN.h
```diff
@@ -71,6 +89,8 @@ class FairMQSocketNN : public FairMQSocket
     std::atomic<unsigned long> fMessagesRx;
 
     static std::atomic<bool> fInterrupted;
+    static bool fHasLingered;
+    static int fLingerDuration;
```
dennisklein: Tracking the linger with global state seems dirty. My counter proposal would be to put `fLingerDuration` as a non-static member and move the actual linger implementation to the `FairMQTransportFactoryNN`. For this, it needs to track the created sockets. On `RESET_DEVICE` we could e.g. call `FairMQTransportFactory::Reset` on all transports, which - in the case of nanomsg - calculates the max of all linger values on all nn sockets and sleeps that long. What do you think?

dennisklein: That leads to another discussion: Should the transport factory give up ownership on the socket. In case of zeromq at least this could break, if one destructs the transport factory before the socket.

fairmq/nanomsg/FairMQTransportFactoryNN.cxx
rbx: Yep, will change

fairmq/nanomsg/FairMQSocketNN.h
dennisklein: > If we do go for owning factories, I would go with unique_ptrs for sockets inside the factory and return non-owning raw ptrs. When device is present we control the lifetime of both factories and sockets, so no additional checks would be necessary at least inside of our classes.

I agree, would do it the same.

dennisklein: Is it ok, if this blocks the other commits in this PR? If you want, you could open a seperate PR just with this commit, so the other ones can be merged already.

rbx: Merging without this will cause the nanomsg test to fail a lot. So I will change the rateLogger to use regular sleep again, until nanomsg is fixed.

rbx: Done, this is now only WaitFor and some changes in the tests.
Nanomsg linger changes will follow in separate PR.

Issue #89, opened 2018-09-17
Option to compile static FairMQ library


mpapenbrock:
I would like to create Docker images my individual FairMQ services, e.g. a sample, a processor, etc. In order to reduce the image size, it would be favourable to compile the service into a static executable which can is then copied into a minimal base image and does not require further dependencies. As far as I am aware, static libraries are needed to achieve that.

An option to compile to compile FairMQ (and by extension FairLogger) into a static library would help to solve this issue.

dennisklein:
I will need to investigate, if this can be done, because FairMQ depends on a number of external libraries which are usually only installed as shared libraries. I will get back to you within the week.

In any case, in order to reduce disk space, make sure you compile FairMQ without debug symbols, e.g. `-DCMAKE_BUILD_TYPE=Release`.

dennisklein:
https://github.com/FairRootGroup/FairLogger/releases/tag/v1.3.0

dennisklein:
Another flag you might want to control is `-fPIC` (on by default). In FairMQ `v1.3.2` you will be able to use `-DBUILD_SHARED_LIBS=OFF -DCMAKE_POSITION_INDEPENDENT_CODE=OFF` to switch it off, i not desired.

dennisklein:
[FairMQ `v1.3.2`](https://github.com/FairRootGroup/FairMQ/releases/tag/v1.3.2) is now available for testing.

mpapenbrock:
I have tested the solution, and I cannot see any issues on the FairMQ side. However, I am getting some undefined references in the static version of the zmq library. I do not believe this has anything to do with FairMQ, and will need some further investigation on my side first.
I think this issue can be closed. Thank you for the effort!

Issue #83, opened 2018-09-11
Deadlock at device shutdown


sawenzel:
**Bug description**

This is a bug observed in AliceO2 DPL workflows. Sometimes FairMQ devices being sent a SIGTERM do not exit properly but go into a waiting state (probably deadlock). We've been seeing this for a while in non-terminating unit tests and it is now becoming a real problem with larger workflows where this occurs more often.

A typical log is the following
```
[66275]: [11:07:42][INFO] Received device shutdown request (signal 15).
[66275]: [11:07:42][INFO] Waiting for graceful device shutdown. Hit Ctrl-C again to abort immediately.
[66275]: [11:07:43][STATE] Entering READY state
** HANGS HERE **
```
which shows that SIGTERM was received but the process does not actually terminate.

At this point, the backtrace is the following:
```
thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x00007fff6ad29a16 libsystem_kernel.dylib`__psynch_cvwait + 10
    frame #1: 0x00007fff6aef2589 libsystem_pthread.dylib`_pthread_cond_wait + 732
    frame #2: 0x00007fff68b2dd43 libc++.1.dylib`std::__1::condition_variable::__do_timed_wait(std::__1::unique_lock<std::__1::mutex>&, std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000000l> > >) + 93
    frame #3: 0x000000010e7e844d libFairMQ.1.3.dylib`std::__1::cv_status std::__1::condition_variable::wait_for<long long, std::__1::ratio<1l, 1000l> >(this=0x00007fe22de1f960, __lk=0x00007ffee5601d78, __d=0x00007ffee5601d70) at __mutex_base:416
    frame #4: 0x000000010e7ec120 libFairMQ.1.3.dylib`fair::mq::fsm::Machine_::ProcessWork(this=0x00007fe22de1f8c0) at FairMQStateMachine.cxx:380
    frame #5: 0x000000010e7ebf7b libFairMQ.1.3.dylib`FairMQStateMachine::ProcessWork(this=0x00007fe22f8e9c00) at FairMQStateMachine.cxx:654
    frame #6: 0x000000010e719e75 libFairMQ.1.3.dylib`FairMQDevice::RunStateMachine(this=0x00007fe22f8e9c00) at FairMQDevice.h:419
[...]

PR #90, opened 2018-09-19
Support static builds

dennisklein:
1. Support CMake variable [`BUILD_SHARED_LIBS`](https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html) (triggered by #89).

2. Fix an ODR violation in the unity build regarding posix signal handling. @rbx Could this be related to the issue with the missed `SIGINT`? Strangely neither of the CI builds generate an error here, only get it with GCC 8.

---

Checklist:

### DIFF
--- a/cmake/FairMQLib.cmake
+++ b/cmake/FairMQLib.cmake
@@ -123,8 +123,14 @@ macro(set_fairmq_defaults)
   endif()
   set(CMAKE_CXX_EXTENSIONS OFF)
 
+  if(NOT BUILD_SHARED_LIBS)
+    set(BUILD_SHARED_LIBS ON CACHE BOOL "Whether to build shared libraries or static archives")
+  endif()
+
   # Set -fPIC as default for all library types
-  set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+  if(NOT CMAKE_POSITION_INDEPENDENT_CODE)
+    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+  endif()
 
   # Generate compile_commands.json file (https://clang.llvm.org/docs/JSONCompilationDatabase.html)
   set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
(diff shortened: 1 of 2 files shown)


rbx:
I'm not aware of any missed signals, what has been reported so far races in handling the signal. Is anyone using unity builds? The Monitor doesn't need to be part of FairMQ library, it is an independent executable. With errors such as ODR violation I would expect to see it either always fail or always happen to work. Cannot imagine how such an issue would contribute to signals being missed sometimes. But again I have never even seen such issues on our side.

dennisklein:
> Is anyone using unity builds?

Yes, alfa-ci.

> The Monitor doesn't need to be part of FairMQ library, it is an independent executable. 

That is how I fixed it, `fairmq-shmmonitor` no longer depends on FairMQ.

dennisklein:
> I'm not aware of any missed signals - what has been reported so far are races in signal handling. 

That is what I meant.

> With errors such as ODR violation I would expect to see it either always fail or always happen to work. Cannot imagine how such an issue would contribute to signals being missed sometimes.

Maybe some strange compiler bug. The facts, that the CI build uses unity build and should have seen the ODR violation, which it did not, and that code regarding the signal handling is affected, made me think this could be related to the observed races which we have not understood yet, haven't we?

rbx:
Right. So far I have only seen the signal issues occur on my mac and as described in #83. Not sure if we have seen them on our test environment, or perhaps we did, but dismissed them as something else?

rbx:
When reproducing #83 on my mac I did not use unity build, I don't think there's any way for the global in the Monitor to affect whatever I was running there, is there?

dennisklein:
Alright, thx for clearing it up. I had to push again, noticed another mistake, didn't mean to delete the PIC setting.

Btw, the analog support for the `BUILD_SHARED_LIBS` flag I added to FairLogger and released it as `v1.3.0`. This also includes now your added mutex.

Issue #54, opened 2018-06-15
Adopt polymorphic memory resources from AliceO2


dennisklein:
@mkrzewic @ktf Action item from today's Framework & Friends meeting.

mkrzewic:
basic integration done: https://github.com/mkrzewic/FairMQ/tree/pmr
now on to tests.

dennisklein:
thx, just the clang format seems broken, if I apply it on my machine, it converts everything back. Which version of clang-format do you use? Are you sure, you use the config from the FairMQ repo?

mkrzewic:
@dennisklein was not there yet :) my version of clang format seems outdated compared to what your config requires, i'll update and see what happens.

mkrzewic:
also: the unit tests are ported and working. Now would be the time to take a step back and think about how to do this better, since up till now I was just porting the O2 implementation. I'm sure a number of things can be done much more elegant directly at the level of fairmq.

mkrzewic:
@dennisklein what version of clang format is needed? the best I can do on centos7 is 5.0 (via llvm-toolset-7), and that seems not to work:
```
YAML:26:16: error: unknown key 'IncludeBlocks'
IncludeBlocks: Regroup
               ^~~~~~~
Error reading /home/mkrzewic/o2/FairMQ/.clang-format: Invalid argument
error: `clang-format -lines=1:169 test/memory_resources/_memory_resources.cxx` failed

```

dennisklein:
`clang-format version 6.0.1 (tags/RELEASE_601/final)`

Then I would say do not apply clang-format and we will do it once it is merged. It is anyways not a strict requirement yet. Most important is, that the code is readable and somewhat consistently formatted, which I have no doubts you will deliver anyways.

> also: the unit tests are ported and working.

Thanks!

> Now would be the time to take a step back and think about how to do this better, since up till now I was just porting the O2 implementation/prototype. I'm sure a number of things can be done much more elegantly directly at the level of fairmq.

Possibly, I will also have a look the next days and give you some feedback latest on Monday.

dennisklein:
Ah, could you please open a `[WIP]` pull request already?

mkrzewic:
ok, i'll apply whatever my clang-format can digest and squash everything together for a WIP PR

dennisklein:
#93 got merged, container adoption topic still pending in #110, needs a bit more time.

PR #92, opened 2018-09-20
[WIP] Adopt FairMQMessage backed memory resource collection from AliceO2

mkrzewic:
As discussed in #54 opening a PR for review. This contains the port of the O2 pmr prototype + basic tools + basic integration into FairMQ + some unit tests (essentially the same as in O2).
Next step in the short term could be additional integration steps (e.g. adding a pointer to factory in FairMQMessage) followed by some simplification of the tools.
In the longer term we will see...


mkrzewic:
wrong branch

PR #97, opened 2018-10-08
Resolve hanging process in case of uncaught exception

dennisklein:
Proposal to solve #95.

@rbx It is not yet working as expected, when I throw in `Run()`:
```
[21:25:27][INFO] DEVICE: Running...
[21:25:27][STATE] Entering ERROR state
terminate called without an active exception
[1]    30405 abort (core dumped)  test/testhelper_runTestDevice --id waitfor_ --control static
```

as opposed to when I throw in `InitTask()` (this behaviour is what I want):
```
[21:24:48][STATE] Entering INITIALIZING_TASK state
[21:24:48][STATE] Entering ERROR state
[21:24:48][DEBUG] Shutting down Plugin Manager
[21:24:48][DEBUG] Unloaded plugin: 'control', version '1.3.3', maintainer 'FairRootGroup <fairroot@gsi.de>', homepage 'https://github.com/FairRootGroup/FairRoot'
[21:24:48][DEBUG] Shutting down Plugin Services
[21:24:48][DEBUG] Destructing device waitfor_
[21:24:48][STATE] Exiting FairMQ state machine
[21:24:48][ERROR] Unhandled exception reached the top of main: init task, application will now exit
```

I also touched a couple of things we can discuss tomorrow.

---

Checklist:

### DIFF
--- a/fairmq/DeviceRunner.cxx
+++ b/fairmq/DeviceRunner.cxx
@@ -111,12 +111,10 @@ auto DeviceRunner::RunWithExceptionHandlers() -> int
     try {
         return Run();
     } catch (std::exception& e) {
-        LOG(error) << "Unhandled exception reached the top of main: " << e.what()
-                   << ", application will now exit";
+        LOG(error) << "Uncaught exception reached the top of DeviceRunner: " << e.what();
         return 1;
     } catch (...) {
-        LOG(error) << "Non-exception instance being thrown. Please make sure you use "
-                      "std::runtime_exception() instead. Application will now exit.";
+        LOG(error) << "Uncaught exception reached the top of DeviceRunner.";
         return 1;
     }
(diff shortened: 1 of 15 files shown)


rbx:
I haven't yet tried to run your code, but I ran into same issue while reproducing/testing the issue.
The terminate happens because in the RunWrapper we start a thread for rate logging and then further in the scope we join it. But since exception happens, it never gets to the join, hence the terminate.

We can replace the rate logging thread with:
```C++
future<void> rateLogger = async(launch::async, &FairMQDevice::LogSocketRates, this);
```
And replace join with:
```C++
rateLogger.get();
```

The `get()` also has the great advantage that it propagates exceptions if they happen inside LogSocketRates, which regular thread does not.

PR #110, opened 2018-10-30
[WIP] pmr container adopt

mkrzewic:
this is to keep the discussion on container adoption open, contains the full code form the original pr #93

### DIFF
--- a/fairmq/MemoryResourceTools.h
+++ b/fairmq/MemoryResourceTools.h
@@ -18,8 +18,6 @@
 namespace fair {
 namespace mq {
 
-using BytePmrAllocator = pmr::polymorphic_allocator<fair::mq::byte>;
-
 //_________________________________________________________________________________________________
 // return the message associated with the container or throw if it is not possible
 template<typename ContainerT>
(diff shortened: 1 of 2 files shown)


fairmq/MemoryResourceTools.h
```diff
@@ -0,0 +1,117 @@
...
+{
+    return std::vector<const ElemT, SpectatorAllocator<const ElemT>>(
+        nelem, SpectatorAllocator<ElemT>(resource));
+};
+
+//_________________________________________________________________________________________________
+/// Return a vector of const ElemT, takes ownership of the message
+template<typename ElemT>
+std::vector<const ElemT, OwningMessageSpectatorAllocator<const ElemT>>
+    getVector(size_t nelem, FairMQMessagePtr message)
```
mattkretz: OK, so this is the intended interface. I'd vote strongly against inclusion of it because:
* The return type name is very long and I expect nobody to be able to spell it out. `auto` can help, but not so much when you want to store it as a data member.
* Users still see `std::vector` in the type and expect it to behave like a vector. I.e. `push_back`, `reserve`, and other operations that reallocate must work correctly. IIUC that's not the case.
* A simple new type with `begin`, `end`, `operator[]`, and `size()` that takes ownership of the `message` and provides an STL interface to read-only access seems easier to create than the current solution (`begin` and `end` could simply return a pointer, no need for a full-blown iterator type) and involves no UB.

mkrzewic: yes, I agree that custom containers that would allow direct adoption would be ideal, standard containers are just not designed to do this. One could try to make it work better (I think it would be possible to get the full expected functionality), but given the fact that gcc8 does not like this (libstdc++ static_asserts that T is not cv) this specific implementation has to go (or grow) anyway. I don't quite understand the need for this assertion as some functionality simply works with const elements, but they decided to start enforcing the requirement for T to be erasable, maybe for consistency.
I still think this is a useful exercise as I like the idea of containers that take ownership of the message together with the data so no additional book keeping is needed along the data path. I'll definitely give it more thought :)
- about the long type names - if we are to use allocators, the type names will be long anyway:)

mattkretz: > I don't quite understand the need for this assertion as **some** functionality simply works with const elements

You already answered your question. ;-)

> if we are to use allocators, the type names will be long anyway

Unless we start using `std::pmr::vector` consistently.

mkrzewic:
here's another go at this still trying to use std::vector - now the adopted vector is not const and fully functional, but still with a special allocator.

dennisklein:
> here's another go at this still trying to use std::vector - now the adopted vector is not const and fully functional, but still with a special allocator.

Thx, will have a look at it soon.

Also found this library: https://github.com/foonathan/array

dennisklein:
So far, I am not convinced that buffer adoption with `std::vector` is something we should do, so closing this. Feel free to reopen, if any new ideas/variations came up.

PR #111, opened 2018-10-30
Allow plugins to create channels

dennisklein:
This also fixes a bug, that prevented the usage of custom types with the
plugin config API.

Example:

```cpp
// in DeviceState::Idle
vector<string> channels;
channels.push_back("name=asdf,method=bind,type=pull");
SetProperty("channel-config", channels);
```

This is a workaround for channel creation from plugins only. One should still configure the channels as usual in `DeviceState::InitializingDevice` via `SetProperty("chans.asdf.0.address", "tcp://...");` (In FairMQ 1.4 channels will be implicitely created, if a plugin sets properties on it).

---

Checklist:

### DIFF
--- a/fairmq/PluginServices.h
+++ b/fairmq/PluginServices.h
@@ -186,13 +186,16 @@ class PluginServices
     auto SetProperty(const std::string& key, T val) -> void
     {
         auto currentState = GetCurrentDeviceState();
-        if (currentState == DeviceState::InitializingDevice)
+        if (   (currentState == DeviceState::InitializingDevice)
+            || ((currentState == DeviceState::Idle) && (key == "channel-config")))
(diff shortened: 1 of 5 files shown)


fairmq/PluginServices.h
```diff
@@ -186,13 +186,16 @@ class PluginServices
     auto SetProperty(const std::string& key, T val) -> void
     {
         auto currentState = GetCurrentDeviceState();
-        if (currentState == DeviceState::InitializingDevice)
+        if (   (currentState == DeviceState::InitializingDevice)
```
rbx:
```suggestion
        if ((currentState == DeviceState::InitializingDevice)
```

PR #118, opened 2018-11-14
Remove column limit from clang format

rbx:
Remove column limit from clang format.

---

Checklist:

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -15,7 +15,6 @@ BinPackArguments: false
 BinPackParameters: false
 BreakBeforeBinaryOperators: NonAssignment
 BreakConstructorInitializers: BeforeComma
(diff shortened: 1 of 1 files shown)


dennisklein:
image why?

rbx:
for beauty :rose:

rbx:
Any half-decent editor will toggle word wrap for you when you need them via key binding. No need to suggest an artificial limit here. This is a silly historical practice that makes code less readable more often than not.

dennisklein:
But it does much more than word wrapping. Too long lines, word wrapped or not, is, what makes code less readable imho. But since we do *not* enforce clang-format (which I vote to keep), you can easily format as you want and even have longer lines. Switching off line width means, I don't get the feature at all.

PR #121, opened 2018-11-27
Update to new DDS 2.2 API

rbx:
- require DDS 2.2
 - fix regressions in automatic port binding
 - fix regression in channel API
 - update DDS example readme

This is a breaking change for the DDS plugin, that requires DDS 2.2 (not backwards-compatible).

---

Checklist:


rbx:
Tests fail because test environments do not have DDS 2.2 yet.
FairSoft needs a patch that bumps DDS to 2.2 to fix this.
But in order to have a working FairSoft with DDS 2.2, FairSoft also needs to bump FairMQ to a version that contains *this* patch :-).

MohammadAlTurany:
Is DDS 2.2 not compatible with older FairMQ?

dennisklein:
We need to finally sort the CI environments. `may18` on Linux is really `dev` and while renaming I should also install the latest dev.

@rbx The circular dependency in FairSoft is just because it contains FairMQ. We should really have separate FairSoft installations for different repos that only contains the dependencies of each repo.

PR #129, opened 2018-12-18
Allow implicit conversion from factory to resource*

mkrzewic:
reduces boilerplate in many places, logic is clear: factory can be treated like a memory resource where appropriate.


mkrzewic:
wrong branch

PR #134, opened 2019-01-29
Remove obsolete dependency to boost signals v1

dennisklein:
Needed to support Boost 1.69, which dropped the signals v1 library. Anyways, we are not using it any more, so we can just remove the dependency.


rbx:
tried this with 1.69 on lxbk - it works.

PR #141, opened 2019-03-11
Extend tests of error cases

rbx:
- test raising SIGINT in every state
 - test going to Error state from every state
 - add new states (bind/connect) to exception tests

 - remove obsolete state machine code

---

Checklist:

### DIFF
--- a/fairmq/StateMachine.cxx
+++ b/fairmq/StateMachine.cxx
@@ -168,7 +168,6 @@ struct Machine_ : public state_machine_def<Machine_>
     Machine_()
         : fLastTransitionResult(true)
         , fNewStatePending(false)
-        , fWorkOngoing(false)
     {}
 
     virtual ~Machine_() {}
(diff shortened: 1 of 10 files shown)


fairmq/StateMachine.cxx
```diff
@@ -464,14 +471,18 @@ void StateMachine::ProcessWork()
         fsm->CallStateChangeCallbacks(State::Idle);
         fsm->ProcessWork();
     } catch(...) {
-        {
-            lock_guard<mutex> lock(fsm->fStateMtx);
-            fsm->fState = State::Error;
-            fsm->CallStateChangeCallbacks(State::Error);
-            fsm->fWorkOngoing = false;
-            fsm->fWorkDoneCV.notify_one();
+        if (fsm->fErrorStateCalledByUser) {
```
dennisklein: Hm, do we enter this catch block twice? Why is this if condition needed?

rbx: Good that you ask, I don't set the condition correctly. This section can be reached either via uncaught exception or by a throw from line 307, so the if checks which one it is. But I don't set the boolean correctly. Will push a fix in a bit. 
... actually a custom exception type (for the case of user transition to error) will solve it better than a boolean, will redo this bit completely.

rbx: Fixed it. If any child of ProcessWork transitions to error state, ErrorStateException will be thrown. If an uncaught exception is thrown within any of the state handlers (`CallStateChangeCallbacks()` or `CallStateHandler()`), it will propagate to catch(...) right away.

PR #146, opened 2019-03-12
Asiofi 0.3.0 and Reconnect loop

dennisklein:


dennisklein:
@rbx I changed two lines around the 100ms sleep. The lock was held for the whole sleep, now it is released just before.

PR #148, opened 2019-03-14
Extending FairMQParts by a constructor taking list of FairMQMessagePtr

matthiasrichter:
This introduces a little helper to create a FairMQParts object in place from a
variable list of arguments. As a side effect also AddParts is extended to support
more than one FairMQMessagePtr.

An example use case is the initialization of a FairMQParts member in a class constructor.

### DIFF
--- a/fairmq/FairMQParts.h
+++ b/fairmq/FairMQParts.h
@@ -31,6 +31,9 @@ class FairMQParts
     FairMQParts(FairMQParts&& p) = default;
     /// Assignment operator
     FairMQParts& operator=(const FairMQParts&) = delete;
+    /// Constructor from argument pack of std::unique_ptr<FairMQMessage> rvalues
+    template <typename... Ts>
+    FairMQParts(Ts&&... messages) : fParts() {AddPart(std::forward<Ts>(messages)...);}
     /// Default destructor
     ~FairMQParts() {};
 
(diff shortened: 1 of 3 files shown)


rbx:
Very nice, thanks!
Could you please also add example usage of this to `examples/multipart/Sampler.cxx`? Just adding a few more parts to the container in there should be fine. The example is also included in the tests, which will make sure your code is tested.

matthiasrichter:
yes, will do. That's a good idea.

rbx:
Ah, wait, the test for that example, checks the number of parts received. So would also need to update the string `"Received message with 2 parts"` in the test definition in CMakeLists.txt of the example.

matthiasrichter:
@rbx, have extended the test. 

Added also one more method to add the content of another `FairMQParts` object by move, also considered in the test. There was some commented code which effectively was a duplicate, now removed.

matthiasrichter:
hmm looks like I need to sort out something on the llvm compilation

rbx:
The test failures are unrelated.

fairmq/FairMQParts.h
```diff
@@ -57,6 +52,23 @@ class FairMQParts
         fParts.push_back(std::move(msg));
     }
 
+    /// Add variable list of parts to the container (move)
+    template <typename... Ts>
+    void AddPart(std::unique_ptr<FairMQMessage>&& first, Ts&&... remaining)
+    {
+        AddPart(std::forward<std::unique_ptr<FairMQMessage>>(first));
```
mattkretz: this needs to be `std::move` instead of `forward`. `first` is an rvalue ref, `remaining` are actual forwarding references.

matthiasrichter: ah true, because it is not a templated parameter, will fix. Thanks for spotting it

rbx:
I pressed 'Update branch' by mistake, which created a merge commit... (in your fork!). @matthiasrichter could you force push a clean version please? sorry, my mistake!

matthiasrichter:
@rbx, no problem, force-pushed my branch again

PR #151, opened 2019-03-21
OFI improvements

dennisklein:
Implements more APIs and applies new `asiofi::semaphore`.


dennisklein:
Codecov.io is still having problems... Merging.

PR #153, opened 2019-03-29
Update state machine picture

rbx:
Checklist:


dennisklein:
Failed tests can be ignored, ports were in use.

rbx:
I wonder if those ports are occupied by other tests from this test run or by devices run by other users on that node. Didn't we serialize the separate tests?

dennisklein:
Not all are serialized. But the machines are used by many users, who knows which ports they use.

PR #159, opened 2019-04-30
Bunch Region Acks when sending over IPC queue

ironMann:
This is performance optimization for releasing of freed SHM messages allocated from a UnmanagedRegion. The change introduce another thread per Region object used to transmit freed region block information in batches. This reduces contention on the IPC synch objects and leads to lower CPU utilization in both, originating and freeing processes.

Also: refactor the `Manager` class since `Region` is no longer copyable or moveable (mutex).

---

Checklist:

### DIFF
--- a/fairmq/shmem/Manager.cxx
+++ b/fairmq/shmem/Manager.cxx
@@ -19,7 +19,7 @@ namespace mq
 namespace shmem
 {
 
-std::unordered_map<uint64_t, Region> Manager::fRegions;
+std::unordered_map<uint64_t, std::unique_ptr<Region>> Manager::fRegions;
 
 Manager::Manager(const string& name, size_t size)
     : fSessionName(name)
@@ -43,7 +43,7 @@ void Manager::Resume()
     // close remote regions before processing new transfers
     for (auto it = fRegions.begin(); it != fRegions.end(); /**/)
     {
-        if (it->second.fRemote)
+        if (it->second->fRemote)
         {
            it = fRegions.erase(it);
         }
@@ -64,11 +64,11 @@ bipc::mapped_region* Manager::CreateRegion(const size_t size, const uint64_t id,
     }
     else
     {
-        auto r = fRegions.emplace(id, Region{*this, id, size, false, callback});
+        auto r = fRegions.emplace(id, fair::mq::tools::make_unique<Region>(*this, id, size, false, callback));
 
-        r.first->second.StartReceivingAcks();
+        r.first->second->StartReceivingAcks();
 
-        return &(r.first->second.fRegion);
+        return &(r.first->second->fRegion);
     }
 }
 
@@ -78,14 +78,14 @@ Region* Manager::GetRemoteRegion(const uint64_t id)
     auto it = fRegions.find(id);
     if (it != fRegions.end())
     {
-        return &(it->second);
+        return it->second.get();
     }
     else
     {
         try
         {
-            auto r = fRegions.emplace(id, Region{*this, id, 0, true, nullptr});
-            return &(r.first->second);
+            auto r = fRegions.emplace(id, fair::mq::tools::make_unique<Region>(*this, id, 0, true, nullptr));
+            return r.first->second.get();
         }
         catch (bipc::interprocess_exception& e)
         {
--- a/fairmq/shmem/Region.cxx
+++ b/fairmq/shmem/Region.cxx
@@ -12,6 +12,8 @@
 
 #include <boost/date_time/posix_time/posix_time.hpp>
 
+#include <chrono>
+
 using namespace std;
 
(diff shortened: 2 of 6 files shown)


fairmq/shmem/Region.cxx
```diff
@@ -49,52 +52,111 @@ Region::Region(Manager& manager, uint64_t id, uint64_t size, bool remote, FairMQ
...
+            blocksToSend = std::min(fBlocksToFree.size(), fAckBunchSize);
+
+            std::copy_n(fBlocksToFree.end() - blocksToSend, blocksToSend, blocks.get());
+            fBlocksToFree.resize(fBlocksToFree.size() - blocksToSend);
+        } // unlock the block mutex here while sending over IPC
+
+        if (blocksToSend > 0)
         {
-            // LOG(debug) << "queue " << fQueueName << " timeout!";
+            while (!fStop && !fQueue->try_send(blocks.get(), blocksToSend * sizeof(RegionBlock), 0))
```
rbx:
```suggestion
            while (!fQueue->try_send(blocks.get(), blocksToSend * sizeof(RegionBlock), 0) && !fStop)
```
Try to send outstanding acknowledgements before checking for the stop signal. This will fix the failing test. The test fails because the code will wait to bunch more than one ack, but our test only transfers a single message. When waiting is done, the code first checks for stop condition that is already true without sending any of the remaining acks. Switching these will first try to send the remaining acks, and then stop.

ironMann: Thanks for looking into this. I see the issue. I changed the behaviour to try to drain the ack queue on exit. Send is will opportunistic in that case since we don't know if the other side is alive to drain the IPC queue.

fairmq/shmem/Region.cxx
```diff
@@ -49,52 +52,118 @@ Region::Region(Manager& manager, uint64_t id, uint64_t size, bool remote, FairMQ
...
 {
-    fWorker = std::thread(&Region::ReceiveAcks, this);
+    fReceiveAcksWorker = std::thread(&Region::ReceiveAcks, this);
 }
 
 void Region::ReceiveAcks()
 {
     unsigned int priority;
     bipc::message_queue::size_type recvdSize;
+    std::unique_ptr<RegionBlock[]> blocks = std::make_unique<RegionBlock[]>(fAckBunchSize);
```
rbx:
```suggestion
    std::unique_ptr<RegionBlock[]> blocks = fair::mq::tools::make_unique<RegionBlock[]>(fAckBunchSize);
```

fairmq/shmem/Manager.cxx
```diff
@@ -64,11 +64,11 @@ bipc::mapped_region* Manager::CreateRegion(const size_t size, const uint64_t id,
     }
     else
     {
-        auto r = fRegions.emplace(id, Region{*this, id, size, false, callback});
+        auto r = fRegions.emplace(id, std::make_unique<Region>(*this, id, size, false, callback));
```
rbx:
```suggestion
        auto r = fRegions.emplace(id, fair::mq::tools::make_unique<Region>(*this, id, size, false, callback));
```

fairmq/shmem/Region.cxx
```diff
@@ -49,52 +52,118 @@ Region::Region(Manager& manager, uint64_t id, uint64_t size, bool remote, FairMQ
...
+    if (fBlocksToFree.size() >= fAckBunchSize)
+    {
+        lock.unlock(); // reduces contention on fBlockLock
+        fBlockSendCV.notify_one();
+    }
+}
+
+void Region::SendAcks()
+{
+    std::unique_ptr<RegionBlock[]> blocks = std::make_unique<RegionBlock[]>(fAckBunchSize);
```
rbx:
```suggestion
    std::unique_ptr<RegionBlock[]> blocks = fair::mq::tools::make_unique<RegionBlock[]>(fAckBunchSize);
```

fairmq/shmem/Manager.cxx
```diff
@@ -78,14 +78,14 @@ Region* Manager::GetRemoteRegion(const uint64_t id)
...
-        return &(it->second);
+        return it->second.get();
     }
     else
     {
         try
         {
-            auto r = fRegions.emplace(id, Region{*this, id, 0, true, nullptr});
-            return &(r.first->second);
+            auto r = fRegions.emplace(id, std::make_unique<Region>(*this, id, 0, true, nullptr));
```
rbx:
```suggestion
            auto r = fRegions.emplace(id, fair::mq::tools::make_unique<Region>(*this, id, 0, true, nullptr));
```

rbx:
Our test environments are still setup with C++11, with the suggested changes all tests should pass and we can merge this.

ironMann:
Suggested changes are in

PR #171, opened 2019-06-25
Require DDS 2.4

dennisklein:
* Remove obsolete FindDDS.cmake
* Adapt to new DDS version

PR #172, opened 2019-06-25
Add include/DDS to path_suffixes

ktf:
For some reason, in our case DDS 2.4 installs in `include/DDS`, not simply `include`.

### DIFF
--- a/cmake/FindDDS.cmake
+++ b/cmake/FindDDS.cmake
@@ -9,7 +9,7 @@
 find_path(DDS_INCLUDE_DIR
   NAMES dds_intercom.h
   HINTS ${DDS_ROOT} $ENV{DDS_ROOT}
-  PATH_SUFFIXES include
+  PATH_SUFFIXES include include/DDS
 )
 
 find_path(DDS_LIBRARY_DIR


dennisklein:
Since we are removing `FindDDS.cmake` in #171 this week it, this fix will be gone.

According to the [release notes](https://github.com/FairRootGroup/DDS/blob/2.4/ReleaseNotes.md) the new header install location is intended. They ship a CMake package now, meaning you could add the suffix to the `INTERFACE_INCLUDE_DIRECTORIES` target property of the dds targets you use. In the case of FairMQ I changed the include statements to the new path because we will not be compatible to 2.2 and below anyways.

ktf:
Could you back port this to 1.4.x branch then? Without it, DDS 2.4 cannot be used there, AFAICT. In alternative, what would you suggest to do for `1.4.x`?

dennisklein:
> Could you back port this to 1.4.x branch then? Without it, DDS 2.4 cannot be used there, AFAICT. In alternative, what would you suggest to do for 1.4.x?

I am in the process of resolving the remaining test issues with #171 which I believe will superseed this proposal. I will double check with the O2 alidist based software stack which will take me another couple of hours, because I don't have a recent compilation lying around. There will be an new testing release v1.4.5 for this (most likely this week).

I also intend to include proper support for boost 1.70 in FairMQ v1.4.5 and check various setups also in concert with your proposal in FairRootGroup/FairRoot#896. I have already started working on 1.70 support on a non-public branch, because the next asiofi release will actually require Boost 1.70, which is needed for converging on the stable FairMQ 1.4 release anyways. So, everything falls together nicely. I just did not expect alidist to move to Boost 1.70 so quickly, but all good.

dennisklein:
Boost requirement/support matrix:

| FairMQ | options | required asiofi | required Boost |
| --- | --- | --- | --- |
| `v1.4.4` | default | not required | `1.64`, but no Boost `1.70` CMake package support |
| `v1.4.4` | `-DBUILD_OFI_TRANSPORT=ON` | `v0.3.1` | `1.66`, but no Boost `1.70` CMake package support |
| `v1.4.5` | default | not required | `1.64`, full support of Boost 1.70 |
| `v1.4.5` | `-DBUILD_OFI_TRANSPORT=ON` | `v0.3.1` | `1.66`, but no Boost `1.70` CMake package support |
| future stable `v1.4.x` | default | not required | probably `1.64`, full support of Boost 1.70 |
| future stable `v1.4.x` |  `-DBUILD_OFI_TRANSPORT=ON` | future `v0.5+` | `1.70` |

So, it is a good thing O2 will already support Boost `1.70` by then. It felt a bit worried to force everyone to Boost 1.70 soonish just because of a feature I need from Boost.Asio - unfortunately, there is no other viable Asio distribution I know of.

PR #177, opened 2019-07-01
Add CMake skeleton for controller SDK

dennisklein:
* New build flag `BUILD_SDK`
* New component `sdk`


dennisklein:
@rbx If there is more functionality to share between core lib and SDK, we can separate it as well. Everything in the fairmq/sdk subdirectory is of course just dummy content, feel free to change as you wish. Unit tests should go together with the other tests into `test`, add a new testsuite and pass the `LINKS SDK` instead of `LINKS FairMQ`.

Checks are green, so this PR is ready for review.

dennisklein:
Let's track the SDK functionality under the `v1.5` milestone for now. If it is in a mature enough state once we do the `v1.4` stable release, we can still add it there.

PR #195, opened 2019-07-31
SDK maintenance

dennisklein:
Working on outstanding issues and quality of new SDK components:

* [x] Adapt to recent API changes in DDS `2.5.24+` (solved by #200)
* [x] Refactor `sdk::Topology::ChangeState` implementation to use asio primitives and normalized types
* [x] Improve doxygen commentary
* ~Introduce `fairmq-top` which polls topology state via `sdk::Topology` and prints a ncurses-based table view similar to `htop`~
   * ~This will be a useful utility but also helps further development of the SDK because it will benchmark the SDK APIs in performance, usability, and correctness.~ *split into a separate PR #203*
* Comply with asio's [AsynchronousOperation concept](https://www.boost.org/doc/libs/develop/doc/html/boost_asio/reference/asynchronous_operations.html) in
  * ~`fairmq-top`~ *split to a separate PR #203*
  * [x] `sdk::Topology::Async*`
  * ~`sdk::DDS*::Async*` (Implement and comply)~ *not needed for now*
* ~#191~ *postponed, do not know a good solution yet*
* ~#192~ *postponed to a later PR*

PR #203, opened 2019-08-28
fairmq-top

dennisklein:
Depends on #195.

Introduce `fairmq-top` which polls topology state via `sdk::Topology` and prints a ncurses-based table view similar to `htop`

This will be a useful utility but also helps further development of the SDK because it will benchmark the SDK APIs in
* performance,
* usability,
* and correctness.


dennisklein:
will be added in other PR

PR #208, opened 2019-09-16
SDK: Replace for loop with count_if

dennisklein:


mattkretz:
👏👏 👍 lgtm

mattkretz:
> I would propose to format it slightly different.

I assume the current formatting is what clang-format produce[ds]? It's certainly a matter of getting familiar with lambda arguments formatted that way. Since clang-format produces this style it was trivial for me to recognize the lambda. I highly recommend to go with what clang-format suggests, since this is becoming the de-facto standard way of formatting C++.

fuhlig1:
I think this depends very much on the style which is used.

rbx:
I think the style in this PR is good, it is also the same as we use everywhere else.
The test failures are unrelated, merging.

PR #209, opened 2019-09-18
Sdk tweaks

rbx:
- Slight optimization in the container handling in the SDK::Topology - simplify the map content and keep the final vector around from the beginning to avoid copy at the end.

- breaking change in the topology xml files: properties that are used for channel communication are now prefixed with `fmqchan_<property_name>`.

### DIFF
--- a/fairmq/sdk/Topology.h
+++ b/fairmq/sdk/Topology.h
@@ -69,6 +69,7 @@ struct DeviceStatus
 };
 
 using TopologyState = std::vector<DeviceStatus>;
+using TopologyStateIndex = std::unordered_map<DDSTask::Id, size_t>; //  task id -> index in the data vector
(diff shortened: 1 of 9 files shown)


fairmq/sdk/Topology.h
```diff
@@ -69,6 +69,7 @@ struct DeviceStatus
 };
 
 using TopologyState = std::vector<DeviceStatus>;
+using TopologyStateIndex = std::unordered_map<DDSTask::Id, size_t>; //  task id -> index in the data vector
```
mattkretz: Do you really need `size_t` here? Do you expect more than 2^31 elements in the data vector? Use `int` to make buckets smaller (I hope), and thus make it more cache efficient.

PR #210, opened 2019-09-18
Use int instead of size_t for the state index container

rbx:
Following suggestion in #209.

### DIFF
--- a/fairmq/sdk/Topology.h
+++ b/fairmq/sdk/Topology.h
@@ -69,7 +69,7 @@ struct DeviceStatus
 };
 
 using TopologyState = std::vector<DeviceStatus>;
-using TopologyStateIndex = std::unordered_map<DDSTask::Id, size_t>; //  task id -> index in the data vector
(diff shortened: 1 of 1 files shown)


fuhlig1:
Wouldn't be an unsigned int better? Or is there a chance that the value can be negative?

mattkretz:
> Wouldn't be an unsigned int better? Or is there a chance that the value can
> be negative?

The first question of signed vs. unsigned is: do you need a number where 
overflow is an error, or do you need modular arithmetic? It probably doesn't 
matter here, but for that reason signed can be optimized better than unsigned.

If 2^31 is not enough range, use 64-bit integers. That STL subscripts 
(index_type) use unsigned is a source of continuous pain, but sadly can't be 
fixed. And for the sake of consistency all new containers and container 
adaptors unfortunately must use size_t as well. (There's a *huge* discussion 
on std::span::size() wanting to be signed... but we can't have nice things.)

dennisklein:
Test failures unrelated and already reported upstream

PR #211, opened 2019-09-27
Add flatbuffers-based command format to SDK/plugin

rbx:
- Add format definitions for the commands that we send between plugins and sdk/controller tools.
- Use the new format in `fair::mq::sdk::Topology`, `fair::mq::plugins::DDS` and `fairmq-dds-command-ui`.

The format is based on flatbuffers, that is added as a private dependency for the new `sdk_commands` component.

This format definitions are much easier to use and extend, compared to the strings that we were sending around. It also easily allows to bundle multiple commands into single buffer/message.

Benchmarking the places where we use this and comparing to old implementation I can see that with flatbuffers the code sections are at least twice as fast.

TODOs:
* [x] Add unit tests.

Solves #192.


CMakeLists.txt
```diff
@@ -343,6 +352,9 @@ if(PROJECT_PACKAGE_DEPENDENCIES)
     elseif(${dep} STREQUAL fmt)
       get_target_property(fmt_include fmt::fmt INTERFACE_INCLUDE_DIRECTORIES)
       get_filename_component(prefix ${fmt_include}/.. ABSOLUTE)
+      elseif(${dep} STREQUAL Flatbuffers)
+        get_target_property(flatbuffers_include flatbuffers::flatbuffers INTERFACE_INCLUDE_DIRECTORIES)
+        get_filename_component(prefix ${flatbuffers_include}/.. ABSOLUTE)
```
dennisklein: Indentation alarm ;)

rbx: ohno

rbx:
> When running with codecoverage flags, something is not working, see https://cdash.gsi.de/viewTest.php?onlyfailed&buildid=215861

Fixed. I was constructing some objects in the wrong order, which happens to work in many cases ([more info](https://github.com/google/flatbuffers/blob/master/include/flatbuffers/flatbuffers.h#L1306)).

The new test fails are unrelated.

PR #213, opened 2019-10-08
Commands: support JSON in addition to binary

rbx:
Allow for the commands to be optionally serialized as JSON instead of binary flatbuffers.
This feature is provided by the flatbuffers implementation.

For this feature flatbuffers needs to see the definitions file at run time (`CommandsFormat.fbs`). To avoid locating and loading this file, I store it to a static string during the build.

Also fix some missing consts.

### DIFF
--- a/fairmq/sdk/commands/Commands.h
+++ b/fairmq/sdk/commands/Commands.h
@@ -27,6 +27,11 @@ namespace sdk
 namespace cmd
 {
 
+enum class Format : int {
+    Binary,
+    JSON
+};
+
 enum class Result : int {
     Ok,
     Failure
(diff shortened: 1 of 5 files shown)


rbx:
Looks like for this to work (specifically for the newly included header `idl.h`) flatbuffers needs to be compiled with `fPIC`, which is not the case in the used FairSoft...

fairmq/sdk/commands/Commands.h
```diff
@@ -314,11 +319,11 @@ struct Cmds
 
     Cmd& At(size_t i) { return *(fCmds.at(i)); }
 
-    size_t Size() { return fCmds.size(); }
+    size_t Size() const { return fCmds.size(); }
     void Reset() { fCmds.clear(); }
 
-    std::string Serialize() const;
-    void Deserialize(const std::string&);
+    std::string Serialize(const FormatType type = FormatType::Binary) const;
```
dennisklein:
```suggestion
    std::string Serialize(Format format = Format::Binary) const;
```

Just a minor comment to the naming: `fair::mq::sdk::cmd::FormatType::Binary` vs `fair::mq::sdk::cmd::Format::Binary`. Format*Type* feels *doppelgemoppelt* to me here.

dennisklein:
I am updating the FairSoft installation on the CI machines

dennisklein:
FairSoft updated, test failures unrelated

PR #217, opened 2019-10-29
Shmem fixes

### LINKED ISSUE #216: Zombie `fairmq-shmmonitor` processes not being reaped?
We have the impression that whenever there is something making `fairmq-shmmonitor` crash, there is nothing which actually handles the SIGCHLD and the processes end up as zombies. Would it be possible to add some sort of reaper logic, or is this something which you expect the device implementor to do?

rbx:
- Shared memory transport: Fix handling of empty message parts. Should fix AliceO2Group/AliceO2#2547, at least partially, still not clear why the empty parts are produced in the first place.
- Extend multipart example/test to include empty parts.
- Fix shmmonitor daemonization. Fixes #216.
- Minor include cleanup, formatting of shmem classes.

### DIFF
--- a/examples/1-1/Sampler.cxx
+++ b/examples/1-1/Sampler.cxx
@@ -6,9 +6,6 @@
  *                  copied verbatim in the file "LICENSE"                       *
  ********************************************************************************/
 
-#include <thread> // this_thread::sleep_for
-#include <chrono>
-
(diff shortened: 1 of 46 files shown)


dennisklein:
Tested with O2 and it appears to fix the segfault.

PR #223, opened 2019-12-04
Update to new DDS API, bump required version

rbx:


rbx:
Updated the test environments with the new DDS.
Some tests involving DDS are still unstable, but it is not related to the changes here, merging.

PR #227, opened 2019-12-16
shmem: properly initialize received multipart messages

rbx:
Fixes an issue that received multipart messages over shmem did not have their meta data fully initialized, which lead to problems when re-sending them.

I will add a test for this case in a separate PR.

### DIFF
--- a/fairmq/shmem/FairMQMessageSHM.h
+++ b/fairmq/shmem/FairMQMessageSHM.h
@@ -21,6 +21,16 @@
 #include <atomic>
 
 class FairMQSocketSHM;
+namespace fair
+{
+namespace mq
+{
+namespace shmem
+{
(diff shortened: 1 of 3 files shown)


fairmq/shmem/FairMQMessageSHM.h
```diff
@@ -9,6 +9,7 @@
 #define FAIRMQMESSAGESHM_H_
 
 #include <fairmq/shmem/Manager.h>
+#include "Common.h"
```
dennisklein: Let's keep this include in the cxx file and fwd declare `MetaHeader` here

rbx: it's an ungly namespaced one, but ok, will have to cleanup namespaces here anyway

PR #229, opened 2020-01-14
Fix issues reported by Coverity

rbx:


fairmq/FairMQChannel.cxx
```diff
@@ -133,26 +133,31 @@ FairMQChannel::FairMQChannel(const FairMQChannel& chan, const string& newName)
...
-    fRateLogging = chan.fRateLogging;
-    fPortRangeMin = chan.fPortRangeMin;
-    fPortRangeMax = chan.fPortRangeMax;
-    fAutoBind = chan.fAutoBind;
-    fIsValid = false;
-    fMultipart = chan.fMultipart;
-    fModified = chan.fModified;
-    fReset = false;
+    {
+        lock_guard<mutex> lock1(fMtx);
```
AnarManafov: Hey @rbx ,

I've just accidentally came across this code today.
I think, it asks for a random deadlock. I am afraid, if these two mutex are used in some other thread or threads in a different order, it will end up with a deadlock.
It might be safer, for example, to use a local temporary object. In this case you will lock one mutex first, then copy, then unlock and lock another one. Or any other algorithm, which will not require locking of two mutex in the same time. It will somewhat slowdown the code, but it's a price to avoid a deadlock in the current design of your class.

rbx: Thanks for the hint! Indeed, a deadlock would be possible here, at least in theory (currently nothing else accesses these in reverse order). 
I would prefer to avoid a temporary object, since this class is relatively heavy.
Looks like this can offer a solution: [std::lock](https://en.cppreference.com/w/cpp/thread/lock).
I'll try it out in a bit.

mattkretz: And consider self-assignment :stuck_out_tongue_winking_eye: 
```cpp
FairMQChannel x{};
x = x; // locks x.fMtx twice
```

rbx: Fixed.

PR #231, opened 2020-01-17
Detect network interface of the default route without use of ip

### LINKED ISSUE #230: Improve implementation of the detection of the default network interface
Currently, the default network interface detection mechanism in FairMQ relies on the `ip`, `grep`, `cut`, and `head` unix tools. Some systems apparently do not provide such tools, so we should improve the implementation to not rely on those.

rbx:
Fixes #230.

detection will first try `/proc/net/route` then `ip route | grep default | cut -d " " -f 5 | head -n 1` and if both fail we will bind on `tcp://*:1`.


rbx:
Test failures are unrelated.

PR #241, opened 2020-02-13
SDK: Support DYLD_LIBRARY_PATH from parent env in `sdk::DDSEnv`

### LINKED ISSUE #235: Support DYLD_LIBRARY_PATH from parent env in `sdk::DDSEnv`
**Describe the bug**
Using of 
```
{
        // Equivalent to calling source DDS_env.sh
        fair::mq::sdk::DDSEnv env;
}
```
on macOS results in runtime error
```
dyld: Library not loaded: libboost_program_options.dylib
  Referenced from: /Users/andrey/DDS/3.1/bin/dds-user-defaults
  Reason: image not found
sh: line 4: 26838 Abort trap: 6           dds-user-defaults --ignore-default-sid -d -c "$HOME/.DDS/DDS.cfg"
```
The problem is that [here](https://github.com/FairRootGroup/FairMQ/blob/19ab8bba3b3c3d59db0e2e75b47dd3b7932421fe/fairmq/sdk/DDSEnvironment.cxx#L50-L65) you execute `dds-user-defaults` by calling `std::system` which internally calls hosts's shell with the parameter command. On macOS `DYLD_LIBRARY_PATH` which you set before in the code will not be propagated to the child shell.

**System information (please complete the following information):**
 - OS: MacOS 10.15.2
 - Compiler: Apple clang version 11.0.0 (clang-1100.0.33.17)
 - Environment: custom installation of DDS, FairMQ, ODC

dennisklein:
Fixes #235

@AndreyLebedev Could you please confirm this resolves #235?

### DIFF
--- a/fairmq/sdk/DDSEnvironment.cxx
+++ b/fairmq/sdk/DDSEnvironment.cxx
@@ -55,12 +55,17 @@ struct DDSEnvironment::Impl
             setenv("HOME", fConfigHome.c_str(), 1);
         }
 
-        std::istringstream cmd;
-        cmd.str("DDS_CFG=`dds-user-defaults --ignore-default-sid -p`\n"
-                "if [ -z \"$DDS_CFG\" ]; then\n"
-                "  mkdir -p \"$HOME/.DDS\"\n"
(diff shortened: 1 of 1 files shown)


AndreyLebedev:
@dennisklein Yes, for me it works now.

PR #249, opened 2020-03-10
Logger and warning fixes

dennisklein:


rbx:
Should also bump the fairlogger requirement to v1.6.2, at least when SDK is built.

rbx:
nevermind, i see what you did there

rbx:
I would rather leave the definition of the `FAIR_NO_LOG`, so we don't force anyone using this header to also use `FAIR_LOG`.

dennisklein:
> I would rather leave the definition of the `FAIR_NO_LOG` to the user, so we don't force anyone using this header to also use `FAIR_LOG`.

Done.

PR #250, opened 2020-03-11
Add example and test of a n-to-m topology, incuding sub-channel use

rbx:


dennisklein:
Test failures unrelated, merging

PR #253, opened 2020-03-30
QC example update

rbx:
- SDK: refactor subscriptions to allow reuse.
- DDSCommandUI: include path argument in ChangeState.
- QC example: rename QC devices, more granular state control in the script.


dennisklein:
Test failures unrelated.

Issue #190, opened 2019-07-26
DDS plugin: Fix unbounded growth of state-changes subscribers if clients fail to unsubscribe


dennisklein:
If a client subscribes to state-changes and fails to unsubscribe (e.g. due to an exception), their subscriptions will never expire which leads to meaningless custom commands sent out with no receiver. We need some way to expire such subscriptions.

dennisklein:
@AnarManafov Is there a way to know for a custom command sender, if their message is internally discarded by DDS, because the recipient no longer exists?

AnarManafov:
@dennisklein , it is a difficult question in terms of implementation.

Look, a custom command passes multiple actors before it reaches the destination: 
Intercom API -> Commander -> Agent -> Intercom API
Since we are in an async environment, at any of this stage the recipient can go away. It will be too complicated for DDS to track the entire chain and confirm whether the command has been delivered or not.  Consider also the fact that the actual CC will be split on on multiple commands as it can have multiple recipients - it makes it even more difficult to track it down.  

On top of that, it is very difficult in general to track whether the command is read from Intercom. Intercom uses a shared memory queue (for multiple reasons) and it is not even possible to detect that the other end is disconnected.
This is why agents automatically drain Intercom queues on task done events.

One thing DDS guarantees is that a given CC will be definitely delivered if the recipient is still subscribed.

I guess, a possible solution would be if a user implements commands with confirmations on top of DDs' CC. 
Also subscribing on CIntercomService::subscribeOnTaskDone will help to track potential recipients who went away.

dennisklein:
@AnarManafov thx for the explanation.

@rbx So, basically, I see two remaining options:
1. Send subscription keepalive messages from the subscriber and implement a timeout on the plugin side.
2. Confirm each change state update and auto unsubscribe on missing confirmation.

Any other idea?

rbx:
>     1. Send subscription keepalive messages (seldomly, like every couple of minutes) from the subscriber and implement a timeout on the plugin side.
> 
>     2. Confirm each change state update and auto unsubscribe on missing confirmation.
> 
> 
> Any other idea?

I think 1. is enough. It is rare that we have more than one subscriber for state changes, and if they politely unsubscribe there is no issue. I feel the amount of traffic 2. would add is not worth it.

dennisklein:
> > ```
> > 1. Send subscription keepalive messages (seldomly, like every couple of minutes) from the subscriber and implement a timeout on the plugin side.
[...]

PR #256, opened 2020-04-10
Sdk subscription heartbeats

rbx:
Solves #190.

When subscribed to state changes SDK will send heartbeats in regular intervals (by default every 60 seconds, configurable) to the devices.

Plugin that receives the heartbeats, will check (lazily, before attempting to send a changed state command) if the last heartbeat from this subscriber arrived less than 3 times the declared interval ago. If so, it will send the command, and if not it will assume subscriber is no longer there and remove it.

### DIFF
--- a/fairmq/plugins/DDS/DDS.cxx
+++ b/fairmq/plugins/DDS/DDS.cxx
@@ -109,16 +109,27 @@ DDS::DDS(const string& name,
                     break;
             }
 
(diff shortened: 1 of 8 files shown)


rbx:
ok, i'll update the interval.

rbx:
updated.

dennisklein:
test failure unrelated

PR #263, opened 2020-05-19
Do not search external GTest by default

dennisklein:
Can be overridden by -DUSE_EXTERNAL_GTEST=ON.

Related to FairRootGroup/FairSoft#250.


dennisklein:
test failures unrelated

PR #266, opened 2020-05-20
Do not handle double SIGTERM as abort

ktf:
A double SIGTERM, or even worse a SIGINT + SIGTERM combination should
not result in an abort, given such a signal really means "die whenever
you want" and can only be generated programmatically. If one wants the
child to die programmatically, they should use SIGKILL.

On the other hand that bashing ctrl-c (i.e. SIGINT) multiple times
on the keyboard really means abort.


ktf:
@Barthelemy

Barthelemy:
that was a fast process, thanks !

rbx:
There is one more pending PR that I would like to have in, once that is in we will release 1.4.17.

PR #267, opened 2020-05-22
Initiate termination process on SIGTERM as well

ktf:
#266 exposed the fact that FairMQ does not really handle SIGTERM. This should make sure processes actually try to quit when received.

### DIFF
--- a/fairmq/plugins/Control.cxx
+++ b/fairmq/plugins/Control.cxx
@@ -23,7 +23,6 @@ namespace
 {
     std::atomic<sig_atomic_t> gLastSignal(0);
     std::atomic<int> gSignalCount(0);
-    std::atomic<bool> gSigtermRequested(false);
 
     extern "C" auto sigint_handler(int signal) -> void
     {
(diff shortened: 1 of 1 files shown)


fairmq/plugins/Control.cxx
```diff
@@ -36,6 +37,8 @@ namespace
 
     extern "C" auto sigterm_handler(int signal) -> void
     {
+        ++gSignalCount;
```
rbx, line 40: I agree on this line. This was missing in #266. But I don't see why the rest is necessary.

ktf: because I was not sure if you wanted to distinguish between term and int. it's fine if we simply increment gSignalCount. Do you want me to change it?

rbx: Yes, unless I am missing something, incrementing the count is sufficient. If you see any other issue, please elaborate. Otherwise update with the increment only. Thanks!

ktf: Ok. Done.

Issue #268, opened 2020-05-24
Disable allow_guessing when parsing boost options


ktf:
We recently introduced a change in DPL which would have removed the "allow_guessing" style for boost::program_options. While this worked fine on our side, it breaks when passing (certain) options to FairMQ, because on its side FairMQ still uses "allow_guessing". We reverted the change (and renamed one of the options to avoid conflicts) however IMHO it would be better if FairMQ disabled the allow_guessing style as well, so that we can disable it on our side (or at least provide an option to do so). For the record the incriminated option is `--config` which conflicts with `--config-key`. Any idea?

dennisklein:
I believe we can disable the *allow_guessing* behaviour. (didn't even know it existed) @rbx any objections?

rbx:
Sounds good to me, let's disable it

ktf:
To disable it in DPL we do the opposite of 

https://github.com/AliceO2Group/AliceO2/pull/3646/files

ktf:
https://github.com/AliceO2Group/AliceO2/pull/3646/files#diff-4de37a90081a5dc6cb4bfa71261f841aL1569-L1581 is the relevant bit.

dennisklein:
available in `v1.4.18` (together with #267)

PR #269, opened 2020-05-25
Adjust program options style to disallow guessing

rbx:
Explicitly list (and override the default) the program option styles. Disable `allow_guessing`.


rbx:
Solves #268. @ktf.

ktf:
Thanks! Are you going to make a tag any time soon?

PR #274, opened 2020-06-03
Add ability to retrieve ZMQ_FD

ktf:


ktf:
I would like to retrieve the underlying socket, to use it with libuv, but so far it does not work (with ipc).

rbx:
I tried your changes to see why it might fail, but it works for me:
```C++
int fd = 0;
size_t fdsize = sizeof(fd);
channel.GetSocket().GetOption("fd", &fd, &fdsize);
LOG(info) << fd;
```
returns 9. Works for both TCP and IPC.

I do this out of InitTask() or Run(). Init() will be too early because there the socket is not yet bound, or even the channel doesn't exist yet.

dennisklein:
If having access to the underlying zmq fd helps you, I am supporting this - but I want to add a "disclaimer": At this point we cannot guarantee any defined behavior for such fds, such as if they edge or level triggered or if they even notify at the right time defined by any concept of the FairMQ API. Especially, we do **not** guarantee that the FairMQ Send()/Receive() will not still block even though the fd has notified.

ktf:
@dennisklein yes, sure.

ktf:
For the record. This + libuv works like charm.

PR #276, opened 2020-06-04
Try to improve session name hashing

ktf:


ktf:
See also discussion in https://alice.its.cern.ch/jira/browse/O2-1485

fairmq/shmem/Common.h
```diff
@@ -107,13 +108,14 @@ struct RegionBlock
...
+// a hash of user id + session id, then we represent it as hex (to increase per
+// char information) and we keep the lower part of the number (because its the one 
+// that varies the most when you have short strings with the same start).
 inline std::string buildShmIdFromSessionIdAndUserId(const std::string& sessionId)
 {
-    boost::hash<std::string> stringHash;
-    std::string shmId(std::to_string(stringHash(std::string((std::to_string(geteuid()) + sessionId)))));
-    shmId.resize(8, '_');
-    return shmId;
+    size_t seed = geteuid();
```
ironMann, line 116: this will require all processes in the chain to be started by the same user. For shm files, we might want to use the group id, to enable chain started by multiple users.

ktf: In that case we should simply force the shmId, no?

rbx: @ironMann The problem comes when users start running topologies with same (default) session id, and end up trying to access same files unintentionally. I fear switching to group id will bring back the same problem, even if less frequently.

@ktf Yeah, I can make shmId configurable, as it is done with the `fairmq-shmmonitor`.

rbx: Setting shmId is not sufficient to share shmem objects between users. Additionally permissions for the underlying file have to be set. It can be done and if it is needed let's open separate issue to support it, it's out of scope of this one.

rbx:
Closing in favor of #277 that replaces the hashing algorithm with SHA2.

PR #277, opened 2020-06-04
Update algorithm to generate hash for shmem id 

rbx:
- Add bundled dependency on PicoSHA2.
- Update hash algorithm that generates hash for shmem id from boost::hash to SHA-2. Former can result in collisions on MacOS, primarily because we have to truncate the hash value to accommodate for the name length limit.

Also:
- Update ZeroMQ transport's GetData() to return `nullptr` for empty messages.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "extern/asio"]
 	path = extern/asio
 	url = https://github.com/chriskohlhoff/asio
+[submodule "extern/PicoSHA2"]
+	path = extern/PicoSHA2
+	url = https://github.com/okdshin/PicoSHA2
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,6 +146,8 @@ if(BUILD_FAIRMQ)
   find_package2(PRIVATE ZeroMQ REQUIRED
     VERSION 4.1.4
   )
+  build_bundled(PicoSHA2 extern/PicoSHA2)
+  find_package2(PRIVATE PicoSHA2 REQUIRED)
 endif()
 
 if(BUILD_TESTING)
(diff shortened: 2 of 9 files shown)


rbx:
This supersedes #276. @ktf

dennisklein:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- test/message/_message.cxx  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/FairRootGroup/FairMQ/pullRequest?prid=5642499&bid=18475095)

davidrohr:
Looks good as far as I can judge, and bundling the sha seems a reasonable approach to me. When can we have a new tag with this + the fix for the alignment?

ktf:
Do we really need a separate package for sha256?

davidrohr:
> Do we really need a separate package for sha256?

If it is bundled? Or would you just copy in a header only version - would also be fine I think.

dennisklein:
> > Do we really need a separate package for sha256?
> 
> If it is bundled? Or would you just copy in a header only version - would also be fine I think.

It is header-only and bundle-only. So from the user's perspective you should not feel it.

davidrohr:
I'd prefer to tag this fast, if you don't mind :)

rbx:
> Can you please add an entry to the top-level COPYRIGHT file, thx

Done.

> I'd prefer to tag this fast, if you don't mind :)

I don't mind.

dennisklein:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- test/message/_message.cxx  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/FairRootGroup/FairMQ/pullRequest?prid=5642499&bid=18475095)

Issue #265, opened 2020-05-19
Add support for custom alignment


rbx:
The message creation interface should accept an alignment parameter that will return the buffer with the requested alignment.

Several scenarios need to be implemented in the transports:
- [x] Shmem Sending - allocate the buffer with the requested alignment, receiver should see the same alignment. #264 adds this part.
- [ ] Shmem receiver - Specific alignment can be requested prior to receiving. Transport should check if the incoming data has the requested alignment, reallocate if it doesn't. Ideally and usually sender/receiver would be synchronized in this regard, so no reallocations are necessary.
- [x] ZeroMQ Sending. ZeroMQ gives no alignment guarantees with the buffer allocated by them. What we can do is allocate our own (aligned) buffer and give it to ZMQ to transfer. This should have no performance penalty.
- [x] ZeroMQ Receiving. If a specific alignment is requested for incoming data, we can check it and reallocate if it doesn't match.
- [ ] Ofi Sending.
- [ ] Ofi Receiving.

If reallocations do happen, they should be added to socket metrics, as `reallocated bytes`/`reallocated msgs`.

rbx:
Closing this. Ofi is out of scope at the moment. And Shmem receiver is not needed as far as I can tell.

PR #282, opened 2020-06-17
Support min alignment in ChannelResource

dennisklein:
Related to #265

### DIFF
--- a/fairmq/MemoryResources.h
+++ b/fairmq/MemoryResources.h
@@ -1,6 +1,6 @@
 /********************************************************************************
  *    Copyright (C) 2018 CERN and copyright holders of ALICE O2                 *
- *    Copyright (C) 2018 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
+ * Copyright (C) 2018-2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
@@ -30,6 +30,7 @@ namespace mq {
 
 using byte = unsigned char;
 namespace pmr = boost::container::pmr;
+using TransportFactory = FairMQTransportFactory;
 
 /// All FairMQ related memory resources need to inherit from this interface
 /// class for the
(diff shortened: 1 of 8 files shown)


fairmq/MemoryResources.h
```diff
@@ -42,70 +43,82 @@ class FairMQMemoryResource : public pmr::memory_resource
...
+auto createMessage(TransportFactory* factory, std::size_t bytes, Alignment alignment) -> MessagePtr;
+
 /// This is the allocator that interfaces to FairMQ memory management. All
 /// allocations are
 /// delegated to FairMQ so standard (e.g. STL) containers can construct their
 /// stuff in
 /// memory regions appropriate for the data channel configuration.
-class ChannelResource : public FairMQMemoryResource
+template<std::size_t MinAlignment>
+class AlignedChannelResource : public FairMQMemoryResource
```
mkrzewic, line 60: is the name change necessary? Since the default behavior does not change (or does it?), signatures do not change (as far as I can see) it seems to be the same animal, just a better behaved one :)

mkrzewic (commented):
How is MinAlignment set? is it something that can be set "globally"? (say once per factory or so?)

dennisklein:
> How is MinAlignment set? is it something that can be set "globally"? (say once per factory or so?)

It would be per factory, but one could create more than one per factory as before, see usage in the unit test in this PR. The `TransportFactory::GetMemoryResource()` is unchanged.

fairmq/MemoryResources.h
dennisklein: If I don't change the name, it would not let me do the name alias: `using ChannelResource = ChannelResource<0>;` - do you know how to make it work?

mkrzewic: indeed, good question :)

mkrzewic:
ok, so it looks essentially like the workaround we have in O2 (@ktf ?). This would give us no further advantage aside from not having this specific allocator in our code base. How about the thing we briefly discussed with @rbx (https://github.com/AliceO2Group/AliceO2/pull/3674#issuecomment-643357655) for a bit more transparency?

fairmq/MemoryResources.h
ktf:
```c++
template <int alignment = 0>
struct ChannelResource {
};
```

no?

dennisklein:
> ok, so it looks essentially like the workaround we have in O2 (@ktf ?). This would give us no further advantage aside from not having this specific allocator in our code base.

Yes, it is nothing more.

> How about the thing we briefly discussed with @rbx ([AliceO2Group/AliceO2#3674 (comment)](https://github.com/AliceO2Group/AliceO2/pull/3674#issuecomment-643357655)) for a bit more transparency?

Sure, I am fine with making the default alignment configurable on that API layer, too.

Just the same question for my understanding, whats the improvement over just using the pmr interface above to create messages? Or, in other words, what makes the pmr solution - even if only a specialized pmr class in O2 repo - just a workaorund? Are we just talking about API design or is there any technical difference you want to see?

mkrzewic:
@dennisklein I'd say it is about API design mostly, technically we have all the pieces, but just would like to avoid confusion. TransportFactory::GetMemoryResource() (and the implicit conversion operator) are just too seductive not to use :) Just saying it could be potentially dangerous if the default API (above) and the thing we actually need (explicitly constructed allocator as it is now) return the same thing different only in subtle ways like (over-)alignment.

Issue #286, opened 2020-07-10
Add UNDEFINED state to fair::mq::sdk::DeviceState enum


AndreyLebedev:
I've already briefly discussed this issue with Alexey. Just want to document it here so that it's not get lost.

**Is your feature request related to a problem? Please describe.**

The idea is to add an UNDEFINED or UNKNOWN or NONE state to fair::mq::sdk::DeviceState enum.
Some use cases:
1) I'm using directly DeviceState enum in my code. What would be a default value to a variable of type DeviceState? Before getting any information from the device the state is technically unknown.
2) The return value from a function when a state can't be retrieved.
3) DeviceState enum is also used to return an aggregated state of the topology. Consider a case when the topology is in a mixed state. There is no single state in this case, for me, the state is also undefined.

dennisklein:
> 1. I'm using directly DeviceState enum in my code. What would be a default value to a variable of type DeviceState? Before getting any information from the device the state is technically unknown.
> 2. The return value from a function when a state can't be retrieved.

:+1: I support adding a "sentinel" value to the `fair::mq::sdk::DeviceState` enum (which is essentially `fair::mq::State`) - I like `fair::mq::State::Undefined`.

> 3. DeviceState enum is also used to return an aggregated state of the topology. Consider a case when the topology is in a mixed state. There is no single state in this case, for me, the state is also undefined.

For this last example, I think, `Undefined` does not fit though. In the [`DeviceState AggregateState(TopologyState)`](https://github.com/FairRootGroup/FairMQ/blob/78b1c188bfb6158a7c1a8db2745a88099dff19da/fairmq/sdk/Topology.h#L103-L114) helper we opted to not map this case at all and instead throw. I see two options, either
1. add a `AggregateTopologyState` enum that also has a value `Mixed` or similar, or
2. - if we want to be fancy - use the optional template, e.g. [`std::optional<DeviceState>`](https://en.cppreference.com/w/cpp/utility/optional) or `boost::optional<DeviceState>` for return values in contexts where exceptions cannot be used and need to be mapped to a scalar.

AndreyLebedev:
[...]

PR #287, opened 2020-07-14
Add Undefined and Mixed states for use in SDK

rbx:
`fair::mq::sdk::AggregateState` no longer throws exception on a mixed state, but returns `fair::mq::DeviceState::Mixed` instead.

Solves #286 @AndreyLebedev.

### DIFF
--- a/fairmq/sdk/commands/Commands.cxx
+++ b/fairmq/sdk/commands/Commands.cxx
@@ -70,9 +70,10 @@ array<string, 17> typeNames =
     }
 };
 
-array<fair::mq::State, 15> fbStateToMQState =
+array<fair::mq::State, 16> fbStateToMQState =
 {
     {
+        fair::mq::State::Undefined,
         fair::mq::State::Ok,
         fair::mq::State::Error,
         fair::mq::State::Idle,
@@ -91,9 +92,10 @@ array<fair::mq::State, 15> fbStateToMQState =
     }
 };
 
-array<sdk::cmd::FBState, 15> mqStateToFBState =
+array<sdk::cmd::FBState, 16> mqStateToFBState =
 {
     {
+        sdk::cmd::FBState_Undefined,
         sdk::cmd::FBState_Ok,
         sdk::cmd::FBState_Error,
         sdk::cmd::FBState_Idle,
--- a/test/sdk/_topology.cxx
+++ b/test/sdk/_topology.cxx
@@ -452,4 +452,26 @@ TEST_F(Topology, SetAndGetProperties)
     ASSERT_EQ(topo.ChangeState(TopologyTransition::ResetDevice).first, std::error_code());
 }
 
+TEST(Topology2, AggregatedTopologyStateComparison)
+{
+    using namespace fair::mq::sdk;
+    ASSERT_TRUE(DeviceState::Undefined == AggregatedTopologyState::Undefined);
+    ASSERT_TRUE(AggregatedTopologyState::Undefined == DeviceState::Undefined);
+    ASSERT_TRUE(DeviceState::Ok == AggregatedTopologyState::Ok);
+    ASSERT_TRUE(DeviceState::Error == AggregatedTopologyState::Error);
+    ASSERT_TRUE(DeviceState::Idle == AggregatedTopologyState::Idle);
+    ASSERT_TRUE(DeviceState::InitializingDevice == AggregatedTopologyState::InitializingDevice);
+    ASSERT_TRUE(DeviceState::Initialized == AggregatedTopologyState::Initialized);
+    ASSERT_TRUE(DeviceState::Binding == AggregatedTopologyState::Binding);
+    ASSERT_TRUE(DeviceState::Bound == AggregatedTopologyState::Bound);
+    ASSERT_TRUE(DeviceState::Connecting == AggregatedTopologyState::Connecting);
+    ASSERT_TRUE(DeviceState::DeviceReady == AggregatedTopologyState::DeviceReady);
+    ASSERT_TRUE(DeviceState::InitializingTask == AggregatedTopologyState::InitializingTask);
+    ASSERT_TRUE(DeviceState::Ready == AggregatedTopologyState::Ready);
+    ASSERT_TRUE(DeviceState::Running == AggregatedTopologyState::Running);
+    ASSERT_TRUE(DeviceState::ResettingTask == AggregatedTopologyState::ResettingTask);
+    ASSERT_TRUE(DeviceState::ResettingDevice == AggregatedTopologyState::ResettingDevice);
+    ASSERT_TRUE(DeviceState::Exiting == AggregatedTopologyState::Exiting);
+}
+
 }   // namespace
--- a/test/sdk/_topology.cxx
+++ b/test/sdk/_topology.cxx
@@ -139,6 +139,32 @@ TEST_F(Topology, ChangeState)
     EXPECT_EQ(sdk::StateEqualsTo(currentState, sdk::DeviceState::InitializingDevice), true);
 }
 
+TEST_F(Topology, MixedState)
+{
+    using namespace fair::mq;
+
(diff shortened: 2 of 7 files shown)


test/sdk/_topology.cxx
```diff
@@ -139,6 +139,32 @@ TEST_F(Topology, ChangeState)
     EXPECT_EQ(sdk::StateEqualsTo(currentState, sdk::DeviceState::InitializingDevice), true);
 }
 
+TEST_F(Topology, MixedState)
```
rbx, line 142: @AndreyLebedev See this test for updated usage. Is it suitable for what you want?

AndreyLebedev: Yes, thanks for the quick implementation!

Can you please make a new tag for FairMQ? And also update FairMQ version in alidist? I will then switch ODC to the newest version of FairMQ.

rbx:
@dennisklein The `States.h` contains a general list of states. The "device" states are chosen in the StateMachine class, which is untouched by this.

dennisklein:
> @dennisklein The `States.h` contains a general list of states. The "device" states are chosen in the StateMachine class, which is untouched by this.

Then at least clarify it please with a comment.

Still, I think the `StateMachine` interface does not reflect what you say, see e.g.:
```cpp
State StateMachine::GetCurrentState() const;
void StateMachine::HandleStates(std::function<void(const State)> callback);
```
To me, this clearly associates the ```State``` type with the set of device states...

rbx:
> > @dennisklein The `States.h` contains a general list of states. The "device" states are chosen in the StateMachine class, which is untouched by this.
> 
> Then at least clarify it please with a comment.
> 
> Still, I think the `StateMachine` interface does not reflect what you say, see e.g.:
> 
> ```c++
> State StateMachine::GetCurrentState() const;
> void StateMachine::HandleStates(std::function<void(const State)> callback);
> ```
> 
> To me, this clearly associates the `State` type with the set of device states...

While that still looks fine to me, the line that actually breaks the abstraction is `using DeviceState = fair::mq::State;`. I'll adjust.

rbx:
I added a separate enum for the aggregated state now, leaving `Mixed` out of `DeviceState`.

AndreyLebedev:
For me, the initial implementation of @rbx looks better. It's more transparent for the user which has to deal only with a single enum. However, I understand @dennisklein concerns regarding the Mixed state. But proper documentation of this special Mixed state would be enough.

In general, I'm fine with both solutions.

rbx:
@AndreyLebedev I renamed `AggregatedState` to `AggregatedTopologyState` and added a comment to clarify the purpose.

AndreyLebedev:
> @AndreyLebedev I renamed `AggregatedState` to `AggregatedTopologyState` and added a comment to clarify the purpose.

Both names are fine for me. I meant your initial implementation where you just add Undefined and Mixed states to fair::mq::State enum.

rbx:
> > @AndreyLebedev I renamed `AggregatedState` to `AggregatedTopologyState` and added a comment to clarify the purpose.
> 
> Both names are fine for me. I meant your initial implementation where you just add Undefined and Mixed states to fair::mq::State enum.

I know :) I was addressing the call for more documentation and the name is part of it.

AndreyLebedev:
> I know :) I was addressing the call for more documentation and the name is part of it.
OK, but for the docs, I meant documenting `Mixed` state. That it's a special case and used only for aggregated topology states.

dennisklein:
> > I know :) I was addressing the call for more documentation and the name is part of it.
> 
> OK, but for the docs, I meant documenting `Mixed` state. That it's a special case and used only for aggregated topology states.

Yes, it only exists in the `AggregatedTopologyStated` enum. That was my whole point - it is a valid state a topology can be in, but not a single device.

fairmq/sdk/commands/Commands.cxx
```diff
@@ -91,9 +92,10 @@ array<fair::mq::State, 15> fbStateToMQState =
     }
 };
 
-array<sdk::cmd::FBState, 15> mqStateToFBState =
+array<sdk::cmd::FBState, 17> mqStateToFBState =
```
dennisklein, line 95:
```suggestion
array<sdk::cmd::FBState, 16> mqStateToFBState =
```

dennisklein:
> > > I know :) I was addressing the call for more documentation and the name is part of it.
> > 
> > 
> > OK, but for the docs, I meant documenting `Mixed` state. That it's a special case and used only for aggregated topology states.
> 
> Yes, mixed shall only exist in the `AggregatedTopologyState` enum. That was my whole point - it is a valid state a topology can be in, but not a single device. I think the PR reflects that now.

There is also another logic error if `State` enum would contain `Mixed`. The `State` enum comes from `<fairmq/States.h>` which belongs to `libFairMQ` our core component. The SDK component `libFairMQ_SDK` depends on the core component. Why should a  header belonging to the core introduce a state that is only meaningful within the optional SDK? The concept of a topology does not exist in the core component at all - so a comment mentioning the SDK would feel out-of-place to me.

dennisklein:
@AndreyLebedev Essentially, wherever you used `State` enum (or the `DeviceState` type alias) before, just use the `AggregatedTopologyState` enum now which contains `Undefined` as well as `Mixed`. And note, we changed the return value of the `AggregateState(TopologyState);` helper and also its behavior to not throw any more when mixed.

Release tag will follow either today or tomorrow.

AndreyLebedev:
> > > > I know :) I was addressing the call for more documentation and the name is part of it.
> > > 
> > > 
> > > OK, but for the docs, I meant documenting `Mixed` state. That it's a special case and used only for aggregated topology states.
> > 
> > 
> > Yes, mixed shall only exist in the `AggregatedTopologyState` enum. That was my whole point - it is a valid state a topology can be in, but not a single device. I think the PR reflects that now.
> 
> There is also another logic error if `State` enum would contain `Mixed`. The `State` enum comes from `<fairmq/States.h>` which belongs to `libFairMQ` our core component. The SDK component `libFairMQ_SDK` depends on the core component. Why should a header belonging to the core introduce a state that is only meaningful within the optional SDK? The concept of a topology does not exist in the core component at all - so a comment mentioning some "SDK aggregated topology state" would feel out-of-place to me.

Honestly, I don't think that a single comment will confuse people. 

Introducing a new enum for me as a user of SDK feels more confusing. Meaning that both enums overlap on 99% with essentially the same logical meaning. For instance, I have a device state. How do I compare it to an aggregated topology state? Technically I can use std::static_cast<int> (which looks bad and only works because of the proper mapping between device and topology states). But logically does a comparison between device state and topology state is a valid operation?

AndreyLebedev:
> @AndreyLebedev Essentially, wherever you used `State` enum (or the `DeviceState` type alias) before, just use the `AggregatedTopologyState` enum now which contains `Undefined` as well as `Mixed`. And note, we changed the return value of the [`AggregateState(TopologyState);`](https://github.com/FairRootGroup/FairMQ/blob/master/fairmq/sdk/Topology.h#L103) helper and also its behavior to not throw any more when mixed.
> 
> Release tag will follow either today or tomorrow.
> 
> p.s.: If you want to compare `DeviceState` with `AggregatedTopologyState` or the other way around just `static_cast<int>`.

Thank you, I've already looked in the implementation.

dennisklein:
> How do I compare it to an aggregated topology state? Technically I can use std::static_cast (which looks bad and only works because of the proper mapping between device and topology states).

I agree that a static_cast does not lead to particularly readable source code, so I added two [operator== overloads](https://github.com/FairRootGroup/FairMQ/pull/287/commits/dfede435255b28e0de6f817a90f13271fc51697f).

> But logically does a comparison between device state and topology state is a valid operation?

Yes, by definition there is a valid comparison of aggregated topology states and device states. (That we use the same name and even the same integer value as of this PR is just implementation detail). However, there is no corresponding device state for the new mixed topo state, of course.

rbx:
@AndreyLebedev This is now included in the FairMQ v1.4.21 release.

PR #288, opened 2020-07-14
Add API to extract ZMQ_EVENTS from socket backend

ktf:
This exposes to the user the ZMQ_EVENTS API, so that we can reliably integrate FairMQ with third party event loops like libuv.
While I appreciate this exposes some internal details of the FairMQSocket, I see no easy solution when introducing multithreading inside a given device as the lack of knowledge about the event types leads to 100% CPU utilisation on the event loop thread (which handles the interaction with fairmq). The single threaded implementation done so far is immune to the issue because the payload itself acts a mitigator.

As for the ZMQ_FD case, this is not expected to a stable API and I am happy to adapt to a better solution which provides the same / similar functionality, when available.

### DIFF
--- a/fairmq/FairMQSocket.h
+++ b/fairmq/FairMQSocket.h
@@ -38,6 +38,9 @@ class FairMQSocket
     virtual void SetOption(const std::string& option, const void* value, size_t valueSize) = 0;
     virtual void GetOption(const std::string& option, void* value, size_t* valueSize) = 0;
 
+    /// If the backend supports it, fills the unsigned integer @a events with the ZMQ_EVENTS value
+    /// DISCLAIMER: this API is experimental and unsupported and might be dropped / refactored in
+    /// the future.
     virtual void Events(uint32_t* events) = 0;
     virtual void SetLinger(const int value) = 0;
     virtual int GetLinger() const = 0;
--- a/fairmq/FairMQSocket.h
+++ b/fairmq/FairMQSocket.h
@@ -38,6 +38,7 @@ class FairMQSocket
     virtual void SetOption(const std::string& option, const void* value, size_t valueSize) = 0;
     virtual void GetOption(const std::string& option, void* value, size_t* valueSize) = 0;
 
+    virtual void Events(uint32_t* events) = 0;
     virtual void SetLinger(const int value) = 0;
     virtual int GetLinger() const = 0;
     virtual void SetSndBufSize(const int value) = 0;
--- a/fairmq/shmem/Socket.h
+++ b/fairmq/shmem/Socket.h
@@ -375,6 +375,15 @@ class Socket final : public fair::mq::Socket
         }
     }
 
+    void Events(uint32_t* events) override
+    {
+        size_t eventsSize = sizeof(uint32_t);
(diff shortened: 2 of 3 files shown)


ktf:
@davidrohr this will be a requirement for the multithreading support in DPL.

dennisklein:
Did you rule out `<FairMQPoller.h>`?

ktf:
No, I've not, but I am not sure how I could use that to integrate with libuv, actually. What I want to achieve is to have the FairMQ event loop and the libuv event loop and task system to coexists. With the current API (without this PR) I have already libuv with epoll / kqueue on FairMQ sockets, which works quite nicely by simply unblocking the uv_run call and giving control back to fairmq. Can I run FairMQPoller in a "one shot" mode?

dennisklein:
It is basically this: http://api.zeromq.org/2-1:zmq-poll. Maybe timeout of 0 is what you mean by "one shot"?

dennisklein:
It is a reactor style interface like epoll / kqueue. You may alternatingly call your epoll and this one - but not sure about fairness then, putting all event fds into a single pollset is probably better.

Off-topic question: Any specific reason why you chose libuv? (In the FairMQ SDK and asiofi I went with the asio event loop in the spirit that this is what is the standard in C++ land, no?)

ktf:
One shot is actually different than "delay = 0". Oneshot means serve the first event and then move on (to the FairMQ state machine loop, in my case).

Yeah, i think the issue is exactly having two of them. I would have to either keep both of them non blocking (but that consumes CPU cycles) or I run one of them in "one shot" mode, but then while it's waiting on the socket for the first event the other one is unserved (unless they know about each other, but I guess zmq_poll knows only about zeromq?). Maybe I could run the poller with delay = 0 and libuv oneshot, but I would still need for libuv to know about zeromq, hence I need this PR.

Regarding boost asio, I do not have particular experience with it, however from my naive eyes libuv seemed (and turned out to be) very easy to embed in both the driver and the devices. I might change it to asio, if there is a strong push for it, but for now libuv does the job.

--
Ciao,
Giulio

dennisklein:
> but then while it's waiting on the socket for the first event the other one is unserved (unless they know about each other, but I guess zmq_poll knows only about zeromq?)

You could probably concurrently poll the two pollsets in two separate OS threads and then notify your worker thread.

Anyways, if @rbx is fine with this, I am fine, too. But may I ask you to add a doxygen comment to the proposed API in this PR that is not stable, thx.

rbx:
I am fine with it under the same disclaimer as Dennis added to your [previous PR](https://github.com/FairRootGroup/FairMQ/pull/274#issuecomment-638237294). Basically somehow documenting it as "unstable" expresses this.

rbx:
Perhaps, as you've done with `ZMQ_FD`, it would suffice to only modify `GetConstant()` and access the property via `Get`/`SetOption()`? Then at least there are no new unstable APIs.

ktf:
@dennisklein I'd rather avoid yet another thread.
@rbx the problem is that while ZMQ_FD is retrieved only once, ZMQ_EVENTS is in the message inner loop, so I am worried about the string->int conversion.

rbx:
@ktf This is now included in the FairMQ v1.4.21 release.

PR #296, opened 2020-08-11
Performance optimizations

ironMann:
- Use relaxed memory ordering to keep the global atomic message counter. Final check uses acquire ordering for read serialization
- Move RateLimit constructor and call to clock::now() out of the fast path. 2x faster message allocation

Screenshot 2020-08-11 at 15 15 33

---

Checklist:

* [ ] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)

### DIFF
--- a/fairmq/shmem/Manager.h
+++ b/fairmq/shmem/Manager.h
@@ -350,8 +350,8 @@ class Manager
         }
     }
 
-    void IncrementMsgCounter() { ++fMsgCounter; }
-    void DecrementMsgCounter() { --fMsgCounter; }
+    void IncrementMsgCounter() { fMsgCounter.fetch_add(1, std::memory_order_relaxed); }
+    void DecrementMsgCounter() { fMsgCounter.fetch_sub(1, std::memory_order_relaxed); }
 
     void SendHeartbeats()
     {


fairmq/shmem/Message.h
```diff
@@ -265,6 +263,7 @@ class Message final : public fair::mq::Message
                 if (fManager.ThrowingOnBadAlloc()) {
                     throw MessageBadAlloc(tools::ToString("shmem: could not create a message of size ", size, ", alignment: ", (alignment != 0) ? std::to_string(alignment) : "default"));
                 }
+                static thread_local tools::RateLimiter rateLimiter(20);
```
mattkretz, line 266: This changes semantics but may be ok: every thread is now limited to 20 `boost::interprocess::bad_alloc` exceptions per second. If there's a longer time where it doesn't throw it may initially throw at a much higher rate than 20Hz to make up for the "missed throws". This may lead to unintended thrashing though.

fairmq/FairMQDevice.cxx
```diff
@@ -507,7 +507,7 @@ void FairMQDevice::RunWrapper()
                 HandleMultipleChannelInput();
             }
         } else {
-            tools::RateLimiter rateLimiter(fRate);
+            static thread_local tools::RateLimiter rateLimiter(fRate);
```
mattkretz, line 510: This changes semantics: The time between `maybe_sleep` calls is measured and determines the new sleep time. It would need an explicit `reset` function to work (which would need to call `clock::now`). But then, why not just construct it on the stack? It should be more efficient on the stack anyway than when it's in global memory (and a new `reset` function gets called).

mattkretz: How about:
```c++
std::optional<tools::RateLimiter> rateLimiter = [&] {
  if (fRate > 0.001) {
    return std::optional<tools::RateLimiter>(std::in_place_t, fRate);
  } else {
    return std::nullopt_t;
  }
}();
// ...
if (rateLimiter) {
  rateLimiter->maybe_sleep();
}
```

ironMann: You're right. Having quick look at `maybe_sleep()` I thought the first invocation might fall through without seeping bc the timestamp from constructor would be too old. However, every subsequent invocation (within the same loop) should do the rate-limiting just fine? Perhaps a stack variable with `0` start time would work better, and `maybe_sleep()` should handle that case (to avoid adding `reset()`)

mattkretz: If the rate limiter ctor doesn't record a time, the first call to `maybe_sleep` has no initial guess for the correct sleep time / back off rate. So it would need an additional state: constructed -> initialized -> rate limiting. Where currently constructed and initialized are combined in the ctor. This optimizes the rate limiter for the cases where it's actually used. It seems you noticed calls to `clock::now` where the rate limiter wasn't actually used, right?

ironMann: That's right, the `clock::now` from constructor was called when rate limiting was not used at all. I meant the initial sleep guess could be derived from the rate parameter itself, and then improved with subsequent `maybe_sleep` calls.

fairmq/shmem/Manager.h
```diff
@@ -430,7 +430,7 @@ class Manager
     std::unordered_map<uint64_t, std::unique_ptr<Region>> fRegions;
 
     std::atomic<bool> fInterrupted;
-    std::atomic<int32_t> fMsgCounter; // TODO: find a better lifetime solution instead of the counter
+    std::atomic<int64_t> fMsgCounter; // TODO: find a better lifetime solution instead of the counter
```
mattkretz, line 433: Why 64-bit? This increases the `sizeof` of this class by 8 bytes (for lock-free atomics) and is potentially a portability issue if anyone ever wants to build for a 32-bit system.

fairmq/shmem/Manager.h
```diff
@@ -148,7 +148,7 @@ class Manager
     void Resume() { fInterrupted.store(false); }
     void Reset()
     {
-        if (fMsgCounter.load() != 0) {
+        if (fMsgCounter.load(std::memory_order_acquire) != 0) {
```
mattkretz, line 151: I don't know the context where `Reset` is used. But from the immediate context the load could also be relaxed, no?

fairmq/FairMQDevice.cxx
mattkretz: But that would still optimize the rate limiter for cases where it's not used. It should be optimized for cases where it's used. (Besides that, the ctor already does a guess from the rate parameter, the first maybe_sleep call then improves on it. It's non-trivial to change & verify how this class works.)
I believe the correct fix is to not use the rate limiter when it's not used. :wink: I.e. don't construct objects you won't need.

fairmq/shmem/Message.h
mattkretz: @rbx Maybe the real question here is how exact you need the rate to be. I guess it's unimportant to reach exactly 20Hz. So how about a simple `this_thread::sleep(50ms)`?

rbx: It doesn't have to be precise. In fact I just recently changed it from `std::this_thread::sleep_for(50ms)` to rateLimiter, with the hope to avoid some sleeps, but I didn't consider the time calculation in the ctor. The change didn't have any impact on my benchmark performance, so I'm fine with changing it back.

mattkretz: Alright. Sounds better than having a `static RateLimiter` here. You're of course right that exception handling is dog slow and thus the sleep might be unnecessary to reach 20Hz. However, the sleep helps to back off and let other threads to their work to free the memory again. (still no fairness, of course)

fairmq/shmem/Manager.h
ironMann: 32bit should be enough, even for our application. Otherwise we're really doing something wrong... There should be 1 instance of this per process, so +8 B is ok even in this economy :)

fairmq/shmem/Manager.h
ironMann: `Reset` should be called only once on process exit, after all `inc/dec` operations have been completed. Default would probably also be fine.

fairmq/shmem/Message.h
rbx: For the allocation method, we can switch back to `std::this_thread::sleep_for(50ms)`. But for the rate limiter in the `RunWrapper` in `FairMQDevice.cxx` precision is more important. There  it is constructed only once per Running state, so the overhead from the constructor is negligible imo.

fairmq/shmem/Manager.h
mattkretz: Right, if it's not performance critical at all, just leave it at sequentially consistent.

fairmq/shmem/Message.h
mattkretz: So the 2x faster message allocation is only due to the rate limiter in Message.h? Is that what you see @ironMann? In that case I'd refrain from changing the code in FairMQDevice.cxx.

ironMann: I only inspected the Message case, in the fig above. Though I see gettime in profiles A LOT, like it's everyone's favorite syscall (why isn't it fetched from the vDSO?) 
I'll drop the ratelimit commit, and @rbx please do find a suitable way to handle the bad_alloc throttling.

mattkretz: I'd love to use a `clock` that doesn't need a syscall. Do you know of any? I mean, in principle, on x86 `rdtscp` should suffice - if you can determine the correct conversion factor to seconds.

ironMann: apparently it's a configuration of glibc: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59177
This really should not be a syscall on linux. Will have to test on Centos8

PR #300, opened 2020-08-25
Shared memory updates

rbx:
- [x] Provide segment debug infos (behind a compile-time option - `FAIRMQ_DEBUG_MODE`, due tue significant performance impact):
  - API: `fair::mq::shmem::Monitor::GetDebugInfo()`.
  - More `fairmq-shmmonitor` output:
    - number of allocated messages
    - print full message list with offsets from segment start.
- [x] Configurable allocation algorithm for managed segment - `--shm-algorithm <rbtree_best_fit/simple_seq_fit>`.
- [x] Option to silence output of `fair::mq::shmem::Monitor::Cleanup()`.
- [x] Add return value to `fair::mq::shmem::Monitor::Cleanup()`, returning a list of objects and if they have been cleaned up or not.
- [x] Initial support for multiple managed segments in the same session - `--shm-segment-id <id>` to specify segment used for allocations in a given device.
- [x] Update the debug tools to support multiple segments - sort message list by segment, etc.

Non-shm changes:
- Remove built-in devices from the main lib.


davidrohr:
@rbx : I have pulled in your branch and it seems to work. I have all processes in NUMA region 0 running with shm-segment-id 0 and in NUMA region 1 with shm-segment-id 1. I see the 0-segment in the fairmq-shmmonitor, and for the workflow in region 1 I don't see anything in the shmmonitor, since it looks for the wrong segment-id. The workflow seems to run stable.

I could not yet test to use the same session for both workflows, for this we have to wait until @ktf prefixes all channels such that the dpl workflows do not collide when running in the same session.

dennisklein (changes requested):
Hm, the `FAIRMQ_DEBUG_MODE` with all those `#ifdef`s looks really scary. Let's bump the minimum required C++ version to C++17 and implement it with `if constexpr` statements. @MohammadAlTurany Do you approve (This means FairMQ 1.4.23+ will require a compiler that supports `-std=c++17`)?

fairmq/shmem/Manager.h
```diff
@@ -95,76 +85,119 @@ class Manager
...
-                fSegments.emplace(0, RBTreeBestFitSegment(create_only, std::string("fmq_" + fShmId + "_main").c_str(), size));
-                LOG(debug) << "created shared memory segment '" << "fmq_" << fShmId << "_main" << "' of " << fSegments.at(0).get_size() << " bytes. Available are " << fSegments.at(0).get_free_memory() << " bytes.";
+                LOG(error) << "something went wrong";
             }
         }
 
         if (mlockSegment) {
             LOG(debug) << "Locking the managed segment memory pages...";
-            if (mlock(fSegments.at(0).get_address(), fSegments.at(0).get_size()) == -1) {
+            if (mlock(boost::apply_visitor(SegmentAddress{}, fSegments.at(0)), boost::apply_visitor(SegmentSize{}, fSegments.at(0))) == -1) {
```
dennisklein:
```suggestion
            if (mlock(std::visit([](auto const& s){ return s.get_address(); }, fSegments.at(0)),
                      std::visit([](auto const& s){ return s.get_size(); }, fSegments.at(0))) == -1) {
```

Just trying out how it would look like with C++17 `std::visit` + lambda. This would also save all those lengthy static_visitor structs.

Or even somethink like this should work:
```suggestion
            if (std::apply(mlock, std::visit([](auto const& s){ return {s.get_address(), s.get_size()}; }, fSegments.at(0))) == -1) {
```

rbx:
Agreed to modernize with C++17. I'll add these changes in a separate PR.

PR #311, opened 2020-09-23
Update return types in ofi::Socket

rbx:
Update return types in ofi::Socket.
alisw/alidist/pull/2494.

@dennisklein we need to include ofi transport in some of the test environments.


rbx:
Tested locally. Merging because trivial.

PR #312, opened 2020-09-29
Channel update

rbx:
- FairMQChannel: Remove methods deprecated since over a year ago.
- FairMQChannel: Remove thread safety. Originally added before we had a dedicated configuration class. At that time channel could have potentially been configured via different entities (but this never occurred in practice). Now all channel configuration goes through the config container and not directly to the channels. Channels are written only once by the device from a single thread. No synchronization in the channel class is needed.
- FairMQChannel: remove unused code.
- FairMQChannel: replace `ResetChannel()` with `Invalidate()`. The former was never fully implemented and not used by anyone (afaik). All that is needed atm is invalidating the channel to force a re-check, if the channel objects are reused between different configuration.
- shmem: fail earlier when given an unsupported socket type.
- small refactoring/cleanup.


dennisklein (approved):
nice

PR #313, opened 2020-10-07
SDK: Remove instance limit from DDSSession

rbx:
It should be possible to construct and use concurrently `sdk::DDSSession` and `sdk::Topology`. The limitations on the DDS side that required this should no longer be there according to @AnarManafov & @AndreyLebedev (can you specify the DDS version since which this is the case? So that we bump the minimum requirement).

- [x] Remove from DDSSession.
- [ ] Remove from DDSEnvironment. Not strictly necessary for the included test case (i think?), but I think we should remove it also.

### DIFF
--- a/fairmq/sdk/DDSSession.cxx
+++ b/fairmq/sdk/DDSSession.cxx
@@ -65,8 +65,6 @@ struct DDSSession::Impl
         , fId(to_string(fSession->create()))
         , fStopOnDestruction(false)
     {
-        setenv("DDS_SESSION_ID", fId.c_str(), 1);
-
         fDDSService.subscribeOnError([](const dds::intercom_api::EErrorCode errorCode, const std::string& msg) {
             std::cerr << "DDS error, error code: " << errorCode << ", error message: " << msg << std::endl;
         });
@@ -81,10 +79,6 @@ struct DDSSession::Impl
         , fStopOnDestruction(false)
     {
         fSession->attach(fId);
-        auto envId(std::getenv("DDS_SESSION_ID"));
-        if (envId != nullptr && std::string(envId) != fId) {
-            setenv("DDS_SESSION_ID", fId.c_str(), 1);
-        }
 
         fDDSService.subscribeOnError([](const dds::intercom_api::EErrorCode errorCode, const std::string& msg) {
             std::cerr << "DDS error, error code: " << errorCode << ", error message: " << msg << std::endl;
@@ -99,11 +93,6 @@ struct DDSSession::Impl
         , fId(to_string(fSession->getSessionID()))
         , fStopOnDestruction(false)
     {
-        auto envId(std::getenv("DDS_SESSION_ID"));
-        if (envId != nullptr && std::string(envId) != fId) {
-            setenv("DDS_SESSION_ID", fId.c_str(), 1);
-        }
-
         // Sanity check
         if (!fSession->IsRunning()) {
             throw std::runtime_error("Given CSession must be running");
@@ -122,10 +111,6 @@ struct DDSSession::Impl
     Impl& operator=(const Impl&) = delete;
     Impl(Impl&&) = delete;
     Impl& operator=(Impl&&) = delete;
-
-    struct Tag {};
-    friend auto operator<<(std::ostream& os, Tag) -> std::ostream& { return os << "DDSSession"; }
-    tools::InstanceLimiter<Tag, 1> fCount;
 
     DDSEnvironment fEnv;
     DDSRMSPlugin fRMSPlugin;
--- a/fairmq/sdk/Topology.h
+++ b/fairmq/sdk/Topology.h
@@ -234,6 +234,8 @@ class BasicTopology : public AsioBase<Executor, Allocator>
         , fDDSTopo(std::move(topo))
         , fStateData()
         , fStateIndex()
+        , fMtx(std::make_unique<std::mutex>())
+        , fStateChangeUnsubscriptionCV(std::make_unique<std::condition_variable>())
         , fHeartbeatsTimer(asio::system_executor())
         , fHeartbeatInterval(600000)
     {
@@ -263,7 +265,7 @@ class BasicTopology : public AsioBase<Executor, Allocator>
     {
         UnsubscribeFromStateChanges();
 
-        std::lock_guard<std::mutex> lk(fMtx);
+        std::lock_guard<std::mutex> lk(*fMtx);
         fDDSSession.UnsubscribeFromCommands();
         try {
             for (auto& op : fChangeStateOps) {
@@ -306,8 +308,8 @@ class BasicTopology : public AsioBase<Executor, Allocator>
         fDDSSession.SendCommand(cmd::Cmds(cmd::make<cmd::UnsubscribeFromStateChange>()).Serialize());
 
         // wait for all tasks to confirm unsubscription
-        std::unique_lock<std::mutex> lk(fMtx);
-        fStateChangeUnsubscriptionCV.wait(lk, [&](){
+        std::unique_lock<std::mutex> lk(*fMtx);
+        fStateChangeUnsubscriptionCV->wait(lk, [&](){
             unsigned int count = std::count_if(fStateIndex.cbegin(), fStateIndex.cend(), [=](const auto& s) {
                 return fStateData.at(s.second).subscribed_to_state_changes == false;
             });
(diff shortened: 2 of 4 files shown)


rbx:
On my machine these changes lead to a problem that I don't yet understand: when running the newly created test all task logs contain only a single line:
```
/tmp/681001b3-a732-4b74-8a08-4a939bbe4d4b/dds_2020-10-07-11-31-58-808/wn/slots/3046755393066700051/dds_user_task_wrapper.sh: line 18: --color: command not found
```
The `--color` is the second task argument from the included topology file, e.g.: `<exe reachable="true">fairmq-bsampler --color false --channel-config name=data,type=push,method=bind -P dds --msg-rate 10</exe>`. It appears that something splits the arguments and tries to run them separately? And idea why @dennisklein @AnarManafov @AndreyLebedev ?

The old tests run properly. I am sure I've seen this issue before, but don't remember how I fixed it. So perhaps it is not related to the new changes.

Because the tasks don't actually start, the test fails when trying to change the state. Let's see if this will also be the case on the test machines.

AndreyLebedev:
@rbx 
Could you please share a topology file?
Also if possible post a content of `/tmp/681001b3-a732-4b74-8a08-4a939bbe4d4b/dds_2020-10-07-11-31-58-808/wn/slots/3046755393066700051/dds_user_task_wrapper.sh` script.

AndreyLebedev:
> The limitations on the DDS side that required this should no longer be there according to @AnarManafov & @AndreyLebedev (can you specify the DDS version since which this is the case? So that we bump the minimum requirement).

Technically it was already possible to create multiple instances of CSession class in v3.0 (which is now required for FairMQ). But there are still some limitations that will be fixed in the coming release of DDS. So if it's feasible for you I propose to bump the DDS version as soon as the next release is available.

rbx:
topo file:
```XML
<topology name="ExampleDDS">

    <property name="fmqchan_data" />

    <declrequirement name="SamplerWorker" type="wnname" value="sampler"/>
    <declrequirement name="SinkWorker" type="wnname" value="sink"/>

    <decltask name="Sampler">
        <exe reachable="true">fairmq-bsampler --color false --channel-config name=data,type=push,method=bind -P dds --msg-rate 10</exe>
        <requirements>
            <name>SamplerWorker</name>
        </requirements>
        <properties>
            <name access="write">fmqchan_data</name>
        </properties>
    </decltask>

    <decltask name="Sink">
        <exe reachable="true">fairmq-sink --color false --channel-config name=data,type=pull,method=connect -P dds</exe>
        <requirements>
            <name>SinkWorker</name>
        </requirements>
        <properties>
            <name access="read">fmqchan_data</name>
        </properties>
    </decltask>

    <declcollection name="Sinks">
        <tasks>
            <name>Sink</name>
        </tasks>
    </declcollection>

    <main name="main">
        <task>Sampler</task>
        <group name="SinkGroup" n="5">
            <collection>Sinks</collection>
        </group>
    </main>

</topology>
```
dds_user_task_wrapper.sh:
```sh
#!/usr/bin/env bash
# Copyright 2019 GSI, Inc. All rights reserved.
#
#

OS=$(uname -s 2>&1)
if [ "$OS" == "Linux" ]; then
   export LD_LIBRARY_PATH=$DDS_LIBS_LOCATION:$LD_LIBRARY_PATH
fi
if [ "$OS" == "Darwin" ]; then
   export DYLD_LIBRARY_PATH=$DDS_LIBS_LOCATION:$DYLD_LIBRARY_PATH
fi

# %DDS_USER_ENVIRONMENT%

# Run user process in background to prevent propogation of the signals
# Otherwise user task will catch the signals sent to the wrapper
# %DDS_USER_TASK% &

# Wait for user task to finish
USERTASK_PID=$!
wait $USERTASK_PID
USER_TASK_EXIT_CODE=$?

exit $USER_TASK_EXIT_CODE
```
The `dds_user_task_wrapper.sh` looks identical to the one in a successful run.

AndreyLebedev:
It seems that DDS could't find `fairmq-bsampler` while parsing the command line. If you don't define an absolute path to the file, DDS tries to find it in a `$PATH`. You can use `<env>` tag to specify the environment script which will set the proper environment.

dennisklein:
@AndreyLebedev Can you please provide a minimal compilable example that runs two CSessions and two custom commands services in a single program (small main.cpp). Based on this example I will remove the FairMQ sdk instance limitations (which only exist because back then DDS provided global custom command interface only which I could not use to connect to two commanders from the same process.) Thx!

rbx:
Adding full logs (including the session dir) for the failed test run. Also one more log from a successfull run that uses same topology file & setup.
[failure.zip](https://github.com/FairRootGroup/FairMQ/files/5340284/failure.zip)
[success.zip](https://github.com/FairRootGroup/FairMQ/files/5340285/success.zip)

AndreyLebedev:
> @AndreyLebedev Can you please provide a minimal compilable example that runs two CSessions and two custom commands services in a single program (small main.cpp). Based on this example I will remove the FairMQ sdk instance limitations (which only exist because back then DDS provided global custom command interface only which I could not use to connect to two commanders from the same process.) Thx!

I'll set up an example soon.
`CIntercomService::start(const std::string& _sessionID = "")` optionally gets a session ID string which identifies to which DDS session to connect. So you can create multiple `CIntercomService` instances and connect them to different DDS sessions.
You can already start adapting the `fairmq::sdk` but some work is also required in DDS in order to make it functioning properly.

rbx:
Fixed the issue with incorrect environment that lead to task executables not being found.

Now the test runs for me, and also the state transitions, done via DDS custom command, succeed to parallel topologies. Now both topologies live at the same time, but the commands are send first to one and then to another. I'll add another test that will send the commands concurrently.

rbx:
Added a concurrent test. Everything seems to be working properly for me locally.

fairmq/sdk/DDSSession.cxx
```diff
@@ -125,7 +125,7 @@ struct DDSSession::Impl
 
     struct Tag {};
     friend auto operator<<(std::ostream& os, Tag) -> std::ostream& { return os << "DDSSession"; }
-    tools::InstanceLimiter<Tag, 1> fCount;
+    // tools::InstanceLimiter<Tag, 1> fCount;
```
dennisklein, line 128:
```suggestion
```

fairmq/sdk/DDSSession.cxx
```diff
@@ -125,7 +125,7 @@ struct DDSSession::Impl
 
     struct Tag {};
```
dennisklein, line 126:
```suggestion
```

fairmq/sdk/Topology.h
```diff
@@ -256,7 +256,21 @@ class BasicTopology : public AsioBase<Executor, Allocator>
...
+        , fStateIndex(std::move(bt.fStateIndex))
+        , fMtx()
+        , fStateChangeUnsubscriptionCV()
+        , fHeartbeatsTimer(std::move(bt.fHeartbeatsTimer))
+        , fHeartbeatInterval(std::move(bt.fHeartbeatInterval))
+        , fChangeStateOps(std::move(bt.fChangeStateOps))
+        , fWaitForStateOps(std::move(bt.fWaitForStateOps))
+        , fSetPropertiesOps(std::move(bt.fSetPropertiesOps))
+        , fGetPropertiesOps(std::move(bt.fGetPropertiesOps))
+    {}
```
dennisklein, line 273: Why?

If it is needed please create seperate commit with commit message containing the rationale.

dennisklein (changes requested):

dennisklein (commented):
In general I don't understand how this PR can already work:

https://github.com/FairRootGroup/FairMQ/blob/aa60c2229048972e92be8ca64fa4363750b6022c/fairmq/sdk/DDSSession.cxx#L64-L68

* How does `fDDSCustomCmd` know to which session it belongs?
* The current implementation sets `DDS_SESSION_ID` process-wide. So, the latest session will win?

rbx:
https://github.com/FairRootGroup/FairMQ/blob/aa60c2229048972e92be8ca64fa4363750b6022c/fairmq/sdk/DDSSession.cxx#L341
This should be sufficient according to @AndreyLebedev, do we then need the env var at all?

fairmq/sdk/Topology.h
rbx: The default move ctor was implicitly deleted because mutex & condition_variable are not movable. Thus the implementation. I moved to a separate commit.

rbx:
I removed the setting of environment variables.

fairmq/sdk/Topology.h
rbx: I now packed both mtx and cv into unique_ptrs and reverted the move ctor back to ` = default`. Feel free to extend this if you want additional checks, or let me know.

dennisklein (approved):

rbx:
Rebased.

Issue #316, opened 2020-10-28
Rework return codes of fairmq-dds-command-ui to return non-zero in error cases


dennisklein:

ChristianTackeGSI:
Especially the "c" (check) command looks like it should have a proper exit code for error states?

ChristianTackeGSI:
Another to consider:
```console
***** fairmq-dds-command-ui -c s
> stop devices
…
Device splitMC-gen cannot transition with STOP
```

rbx:
> Another to consider:
> 
> ```
> ***** fairmq-dds-command-ui -c s
> > stop devices
> …
> Device splitMC-gen cannot transition with STOP
> ```

Which FairMQ tag produces the ` ... cannot transition with ... ` message?

ChristianTackeGSI:
> Which FairMQ tag produces the `... cannot transition with ...` message?

Probably the old one from FairSoft:spack:jun19.
So I guess, you can ignore it.

I assume, Dennis and you know very well, which runtime paths should return an exit code != 0.

The main rule is: Give shell scripts an easy way to react to an error, if the error is severe enough.
If error scenarios are complex enough, you could even opt for different exit codes for different error classes. But I'd only go that route, if you can see clear ways of shell scripts having to react differently. Otherwise: Keep it simple.

rbx:
So, I added the handling for the errors that SDK reports. See the linked PR.
 
> Especially the "c" (check) command looks like it should have a proper exit code for error states?

Thinking about this particular case, I'm not sure. This command requests states of all devices. If some of the devices are in the error state, is that a failure condition for the dds-ui-tool? Or is it's job simply to report the situation, and the consumer of this data should decide further. @dennisklein what do you think? I'm not strongly for or against it, but compared to other error conditions this one gives me a pause.

ChristianTackeGSI:
A few thoughts about "c" check and exit codes:

If "the consumer" should be able to do decide, then you might consider to make the output "machine readable" and "stanle" (and declare that property officially). "stable" means, that there should not be any changes because of "better wording", or "formatting".

One obvious answer: If all (selected?) states are ERROR, then the exit code should indicate that.

For only some being in ERROR state: You're right, that case is not simple. I don't have a simple answer there.

rbx:
Thinking about it more, it's better to return non-zero even if any of the devices are in error. The consumer just has to take that case into account. Updated the PR.

Issue #303, opened 2020-09-02
`fairmq-dds-command-ui -c c` always shows `UNDEFINED`


dennisklein:
We just found with @karabowi that the current implementation of the `-c c` command does not wait for the `fair::mq::sdk::Topology`'s internal state subscription to succeed and thus reports the `UNDEFINED` values which are set at initialization.

https://github.com/FairRootGroup/FairMQ/blob/690e8a037098ed4b7097f687d853d20a2b8cc028/fairmq/plugins/DDS/runDDSCommandUI.cxx#L61-L66

dennisklein:
I would avoid to make `fair::mq::sdk::Topology::GetCurrentState()` blocking.

@rbx We could add a ctor argument `bool blockUntilConnected` which blocks the ctor until subscription acks have been received. Or as a separate member function `WaitUntilConnected()`. What do you think?

rbx:
> I would avoid to make `fair::mq::sdk::Topology::GetCurrentState()` blocking.
Agreed, it returns only local state by design.
> @rbx We could add a ctor argument `bool blockUntilConnected` which blocks the ctor until subscription acks have been received. Or as a separate member function `WaitUntilConnected()`. What do you think?
I'm fine with both/any. 

But what you can already do now is:
https://github.com/FairRootGroup/FairMQ/blob/master/examples/dds/fairmq-start-ex-dds.sh.in#L55
Basically in the beginning wait until devices report to be IDLE.
That's why I wouldn't classify this issue as a bug.

PR #317, opened 2020-11-11
Modify DDS Command UI return codes

rbx:
- SDK Topology: Add `WaitForPublisherCount(num)` to wait for specified number of tasks to report their states. Add related ctor argument. I made the `SDK::Topology` default to not wait (behaves as before), but for `fairmq-dds-command-ui` the default is to wait. Solves #303. 
 - DDS Command UI: return `EXIT_FAILURE` on errors. Solves #316.
 - Apply suggestions from #318. Solves #318.


fairmq/sdk/Topology.h
```diff
@@ -308,13 +321,7 @@ class BasicTopology : public AsioBase<Executor, Allocator>
         fDDSSession.SendCommand(cmd::Cmds(cmd::make<cmd::UnsubscribeFromStateChange>()).Serialize());
 
         // wait for all tasks to confirm unsubscription
-        std::unique_lock<std::mutex> lk(*fMtx);
-        fStateChangeUnsubscriptionCV->wait(lk, [&](){
-            unsigned int count = std::count_if(fStateIndex.cbegin(), fStateIndex.cend(), [=](const auto& s) {
-                return fStateData.at(s.second).subscribed_to_state_changes == false;
```
rbx, line 314: I think with my changes here, we don't need the `subscribed_to_state_changes` member at all. But since it is part of the public API, I leave it in. Perhaps someone will even find it useful to have this info. Also, this member feels like an implementation detail to me.

dennisklein: Why did you add the counter?

rbx: It's faster and easier to check the counter when waiting for all to subscribe or unsubscribe, than to do a count_if each time the count changes. It's used both in ctor and dtor.

rbx: Your question makes me think, what if a task erroneously sends a un/subscription message more than once. Then counter will be wrong as it is implemented now. I will add a check and warning in the SDK if that happens.

dennisklein: I recommend to factor the count_if stuff into its own function and then use a counter as a "cache" type of thing. would also need a counterpart function to update the subscription status.

rbx: Why? with the counter there is no need for count_if at all.

dennisklein: Well, on the getter side, no, but you would still need `fStateData` to get your counter to count the right thing. So, the logic (that accesses the `fStateData`) would move to the setter side, right?

dennisklein: If I understand your counter correctly, your argument is: When we need to query the amount of subscribers, let's re-implement it by not counting on every query, but by pre-populating the query result in a variable which we update on every update on the subscribers, right?

dennisklein: standard space-time trade-off thingy

rbx: yes

dennisklein (changes requested):
Hm, should we catch the errors in the interactive loop still? I think aborting there would be undesirable?

rbx:
Right! Updated.

dennisklein:
This one looks new: https://cdash.gsi.de/testDetails.php?test=7407194&build=252516 @rbx can you investigate? The dds logs are in https://alfa-ci.gsi.de/job/FairRootGroup/job/FairMQ/job/PR-317/4/artifact/codecov_Debian8-x86_64-gcc9.1.0-FairSoft_fairmq_dev_dds_logs.tar.gz
Note: this ran with `CMAKE_BUILD_TYPE=Profile` not sure if it makes a difference.

rbx:
Interesting, will check.

rbx:
It looks like custom commands from one session are sent to another. @AndreyLebedev can you take a look at the DDS logs Dennis has linked? Sessions `550d6cdf-1a01-4d85-860a-ae4bfded1e26` and `b45a0997-77e4-443e-9448-4ff373bbdeeb` are the ones in question.

rbx:
@dennisklein Is it DDS 3.0? Don't we need a later tag in order to be able to have multiple sessions in one process?

dennisklein:
> @dennisklein Is it DDS 3.0? Don't we need a later tag in order to be able to have multiple sessions in one process?

It is DDS 3.0, see https://cdash.gsi.de/buildSummary.php?buildid=252516

dennisklein:
> @dennisklein Is it DDS 3.0? Don't we need a later tag in order to be able to have multiple sessions in one process?

If higher DDS version is needed, either guard this unit test with a check against `DDS_VERSION` or require it globally in the `find_package2` statement.

rbx:
That's right, we didn't bump the version when adding those tests because there was no tag available at that time. @AndreyLebedev is there now a tag that has full support for multiple sessions in one process?

AndreyLebedev:
> That's right, we didn't bump the version when adding those tests because there was no tag available at that time. @AndreyLebedev is there now a tag that has full support for multiple sessions in one process?

You need version `3.5.3` of DDS which fully supports multiple sessions in a single process.

rbx:
I bumped the FairMQ DDS requirement to 3.5.3. However this will probably mean that none of our test machines can build it now. Anything I can do help update them?

dennisklein (approved):

PR #321, opened 2020-11-17
Let single message methods also return int64_t

rbx:
Due to historical reasons (ZeroMQ transfer methods return int) all single-message transfer methods return `int`. While this is fine for ZeroMQ, other transports are capable of transferring larger data sizes in a single message. This PR changes all transfer methods return types to int64_t.

Triggered by AliceO2Group/AliceO2/pull/4804 (although not the cause for any trouble in that case).


dennisklein (approved):

dennisklein:
Test failure is not related. macOS 10.14 builds are unavailable from now on, will update the CI config separately.

PR #323, opened 2020-11-25
shm: ensure local segments are complete for region subscriptions

rbx:
Fixes the case when region subscriptions locate existing shared memory segment that does not yet have corresponding transport-local metadata initialized (fixed by retrieving the missing data in such a case).


rbx:
Test failures unrelated. Merging.

PR #325, opened 2020-12-04
shm: do mlock/zeroing under shmem lock

rbx:
If mlocking/zeroing of the segment is done outside of the shmem lock, it is possible that region event consumers may be notified before segment is locked/zeroed, which is undesirable in some cases. These changes fix it by doing mlock/zeroing under the same shmem lock.


rbx:
Test failures unrelated. Merging.

PR #329, opened 2021-02-09
shmmonitor: non-interactive mode checks and quits

rbx:
With this change the `fairmq-shmmonitor` (when run with `-v` (view only)) will output the segment contents and exit immediately.
To get continuous output, use `-i` (interactive).
The output format is same for both.

@sawenzel This does what you requested.
@davidrohr FYI: with this you might have to modify your options if you have been using `-v` only, but want continuous output.

### DIFF
--- a/fairmq/shmem/Monitor.cxx
+++ b/fairmq/shmem/Monitor.cxx
@@ -45,6 +45,27 @@ namespace
 namespace fair::mq::shmem
 {
 
+struct TerminalConfig
+{
+    TerminalConfig()
+    {
+        termios t;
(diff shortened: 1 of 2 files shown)


davidrohr (approved):
thx, actually I wanted to request that feature myself :) +1

sawenzel (approved):

sawenzel:
Thank you!

sawenzel:
@rbx : Can you let us know when this is released?

rbx:
@davidrohr @sawenzel Released with v1.4.30

PR #333, opened 2021-02-20
RFC: cmake: Use `target_compile_features()`

ChristianTackeGSI:
Problem at hand: dependents (like FairRoot) need to know the minimum C++ standard level that FairMQ (and its headers) requires.

The first idea is to let the targets export their `CXX_STANDARD` value. First, this doesn't seem to work as expected.

Second, `target_compile_features()` seems to be the better way to go. It has a much better granularity, automatically has the export feature, and thus should make dependents behave correctly.

Also drop all of this enforeced CMAKE_CXX_STANDARD* setting. If it's given, check it. But that's it.

See: https://gitlab.kitware.com/cmake/cmake/-/issues/18446
See: https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#requiring-language-standards

* [X] Rebased against `dev` branch
* [X] My name is in the resp. CONTRIBUTORS/AUTHORS file
* [X] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)


ChristianTackeGSI:
This is not necassarily the final or good version of this.

This is a version, that works for me.

And I want to start some discussion on this.

cmake/FairMQLib.cmake
```diff
@@ -94,10 +94,8 @@ macro(set_fairmq_defaults)
   endif()
 
   # Handle C++ standard level
-  set(CMAKE_CXX_STANDARD_REQUIRED ON)
-  if(NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD ${PROJECT_MIN_CXX_STANDARD})
-  elseif(${CMAKE_CXX_STANDARD} LESS ${PROJECT_MIN_CXX_STANDARD})
+  if(DEFINED CMAKE_CXX_STANDARD AND (NOT "${CMAKE_CXX_STANDARD}" STREQUAL "")
+     AND "${CMAKE_CXX_STANDARD}" LESS ${PROJECT_MIN_CXX_STANDARD})
```
dennisklein, line 98:
```suggestion
  if(CMAKE_CXX_STANDARD AND CMAKE_CXX_STANDARD VERSION_LESS PROJECT_MIN_CXX_STANDARD)
```

dennisklein (commented):
Looks fine. Need to check how this behaves with FairRoot and AliceO2 (in general projects that set `CMAKE_CXX_STANDARD` in addition) before merging.

ChristianTackeGSI:
> Need to check how this behaves with FairRoot

FairRoot has a bit questionable setup, because it does not set `*CXX_STANDARD ` at all.:
```console
FairRoot$ grep CXX_STANDARD -r .
FairRoot$
```

In contrast, it has code that forces people to pass in `-std=` flags by hand, so that cmake has no chance of analyzing them correctly:
https://github.com/FairRootGroup/FairRoot/blob/dev/CMakeLists.txt#L115-L124

This yields to the manually passed flag and the flag created by cmake to both end up in the final compiler call.

I will look into this deeper tomorrow.

> and AliceO2 (in general projects that set CMAKE_CXX_STANDARD in addition)

Who could we ping to check this for AliceO2?

My general hope is, that cmake just does the right thing when a target has `CXX_STANDARD` specified (via CMAKE_CXX_STANDARD) and in addition an imported target also has compiler_features set. I would assume, that it just switches the compiler to a mode that will satisfy everything.

dennisklein (approved):
will merge once I fixed the CI

dennisklein:
cherry-picked in #363

PR #343, opened 2021-03-17
Shm no cleanup option

rbx:
- add --shm-no-cleanup option. When true, device will skip the segment cleanup even when it is the last segment user.
- shm: Improve error message when segment cannot be opened.


rbx:
@davidrohr this option should do what we discussed a while ago.

PR #345, opened 2021-03-23
CI: Run clang-tidy

dennisklein:


dennisklein:
> Does this really make sense? Since those files are not in the same directory. Maybe use the full path instead (<fairmq/etc>)?

Ya, I will probably move more bits around until I am happy.

rbx:
Let's merge this? I want to test the checks already 😆

rbx (approved):

PR #349, opened 2021-04-08
Fallback to <boost/filesystem> on GCC 7

dennisklein:
see alisw/alidist#2949


rbx (approved):

dennisklein:
Something is still off with the locale in the CI container
```
locale::facet::_S_create_c_locale name not valid
```
Merging for now.

PR #354, opened 2021-05-06
Thread local cache for Region lookup

ironMann:
Improve lookup and cache locality of shm Manager.

---

Checklist:

### DIFF
--- a/fairmq/shmem/Manager.h
+++ b/fairmq/shmem/Manager.h
@@ -314,6 +314,7 @@ class Manager
 
                 (fEventCounter->fCount)++;
             }
+            fRegionsGen += 1; // signal TL cache invalidation
             fRegionEventsCV.notify_all();
 
             return result;
@@ -327,8 +328,28 @@ class Manager
 
     Region* GetRegion(const uint16_t id)
     {
+        // NOTE: gcc optimizations. Prevent loading tls addresses many times in the fast path
+        const auto &lTlCache = fTlRegionCache;
+        const auto &lTlCacheVec = lTlCache.fRegionsTLCache;
+        const auto lTlCacheGen = lTlCache.fRegionsTLCacheGen;
+
+        // fast path
+        for (const auto &lRegion : lTlCacheVec) {
+            if ((lRegion.second == id) && (lTlCacheGen == fRegionsGen)) {
+                return lRegion.first;
+            }
+        }
+
         boost::interprocess::scoped_lock<boost::interprocess::named_mutex> lock(fShmMtx);
-        return GetRegionUnsafe(id);
+        // slow path: check invalidation
+        if (lTlCacheGen != fRegionsGen) {
+            fTlRegionCache.fRegionsTLCache.clear();
+        }
+
+        auto *lRegion = GetRegionUnsafe(id);
+        fTlRegionCache.fRegionsTLCache.emplace_back(std::make_pair(lRegion, id));
+        fTlRegionCache.fRegionsTLCacheGen = fRegionsGen;
+        return lRegion;
     }
 
     Region* GetRegionUnsafe(const uint16_t id)
@@ -366,6 +387,7 @@ class Manager
             fShmRegions->at(id).fDestroyed = true;
             (fEventCounter->fCount)++;
         }
+        fRegionsGen += 1; // signal TL cache invalidation
         fRegionEventsCV.notify_all();
     }
 
@@ -613,6 +635,7 @@ class Manager
         using namespace boost::interprocess;
         bool lastRemoved = false;
 
+        fRegionsGen += 1; // signal TL cache invalidation
         UnsubscribeFromRegionEvents();
 
         {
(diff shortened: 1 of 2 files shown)


ironMann:
I switched to a vector lookup with only a handful of elements it's faster than the map. Gcc does not optimize references for TLS so the implementation is a bit ugly (taking local refs first). Fast path avoids touching the "Generation" atomic unless it finds the match first.

rbx:
I have moved the message counter to debug mode completely in my private branch, that is ready to be merged too.
We can of course still keep it optimized even in debug mode, but I would rather keep it simple there. Default build will not have it at all.

ironMann:
Good, that counter will not be missed ;)

rbx:
Meanwhile I found another problem, which I think is still present in this version.
If two sessions(=shmids) are created in the same thread and both create regions, there will be a clash. Region in each session will have id 1 and since thread is the same, the cache will not be able to differentiate. Possible solution would be to add shmid (uint64) to the vector entry (e.g. tuple<ptr, region_id, shm_id>). 

Although this is not relevant in O2 case, we need to make sure this works for the use-as-library use case.

fairmq/shmem/Manager.h
```diff
@@ -665,9 +693,14 @@ class Manager
     Uint16SegmentInfoHashMap* fShmSegments;
     Uint16RegionInfoHashMap* fShmRegions;
     std::unordered_map<uint16_t, std::unique_ptr<Region>> fRegions;
+    // make sure this is alone in the cache line: mostly read
+    alignas(128) inline static std::atomic<unsigned long> fRegionsGen = 0ul;
```
rbx, line 697: Does the `alignas` really do what the comment says? It aligns the long, but since there is no padding after, something can land after it in the cache line?

rbx:
Could you drop the merge commit please (`git reset 53786030d53a77c9ff2e62c1b7de795355621310` should do it).

fairmq/shmem/Manager.h
ironMann: No, you're right. Padding must be manually added.

rbx: Thanks to @mattkretz for noticing.

mattkretz: Late to the party. But I wanted to point out the interesting difference of declaring a member with overalignment and declaring a type with overalignment. The latter also affects the `sizeof` of the type (adding padding), so that a C-array of the type has the correct alignment for all entries. The former also makes the sizeof of the enclosing type a multiple of the largest alignment but *not* the overaligned member.
```c++
struct alignas(64) A {};
struct B {
    alignas(64) int a;
};

// 64 == sizeof(A);
// 64 == sizeof(B);
//  4 == sizeof(B::a);
```

ironMann:
would do it, but git push just blocks forever... just for this repo?!

fairmq/shmem/Manager.h
mattkretz: Oh, and overalignment can potentially waste a lot of memory. If possible (basically only construction order and ABI compatibility concerns), reorder the members from largest alignment to smallest alignment.

rbx:
> would do it, but git push just blocks forever... just for this repo?!

Strange, must be just a hickup. You should be pushing to the same repo as before (to your fork).

fairmq/shmem/Manager.h
ironMann: We have two use cases for over alignment:  
1 nasty counters that keep bouncing the cache line when changed from multiple threads; isolate from the rest of data
2 stable and often used values that could happily live in many caches in the shared state: prevent neighboring variables to invalidate the cache line.

Memory wasting is not usually the worry. Cache line should be mostly 64B, but I remember Intel prefetchers always working with aligned 128B of data (2 CL).

ironMann:
yes, ok, had to add ssh remote... should be there now

fairmq/shmem/Manager.h
mattkretz: Oh, I missed that `fRegionsGen` is `static`. I.e. it doesn't affect the `sizeof` (and padding) of this structure at all. But AFAIK, it still doesn't ensure that other global objects don't go into the same cacheline (behind it).

ironMann: Yes, strange rules, but prob for a good reason. Unfortunately, `alignas` does not work with typedefs to change the `sizeof` of existing types. It has to be a struct or union...

mattkretz: I don't find it strange, after all you're just asking for a certain alignment of the variable, not for any padding. The compiler might need to add padding before the object to achieve the alignment, but that's it. If you require padding after the atomic you must use a structure. E.g.:
```c++
struct A {
    inline static struct alignas(128) {
        int x;
    } x = {0};
};
```
https://godbolt.org/z/nnb9PPoPo

dennisklein:
Took the liberty to rebase this against `dev` (which now contains everything from the `v1.4.35_hotfix_alice2` branch).

rbx:
@ironMann Do you still want to add something in this PR?

ironMann:
LGTM, ship it !

PR #358, opened 2021-05-11
GCC11 support

dennisklein:


dennisklein:
ups, already had a PR open, continued in #363

PR #359, opened 2021-05-13
Reduce noise in examples

rbx:
- Remove `const` in `getDevice` from `runDevice.h`.
- Simplify examples file structure.


dennisklein (approved):
Not going to read it :P

PR #363, opened 2021-05-26
GCC11 Fedora

dennisklein:


CMakeLists.txt
```diff
@@ -16,13 +16,25 @@ include(FairMQLib)
...
+set(PROJECT_FairCMakeModules_VERSION 0.2)
+find_package(FairCMakeModules ${PROJECT_FairCMakeModules_VERSION} QUIET)
+if(NOT FairCMakeModules_FOUND)
+  build_bundled(FairCMakeModules extern/FairCMakeModules)
+  find_package(FairCMakeModules REQUIRED)
+endif()
+
+include(FairFindPackage2)
+include(FairFormattedOutput)
+include(FairSummary)
```
rbx, line 32: This is included twice? here and

CMakeLists.txt
```diff
@@ -251,99 +247,11 @@ install_cmake_package()
...
-    endif()
-    if(NOT ${dep}_BUNDLED)
-      message(STATUS "  ${BWhite}${dep_padded}${CR}${version_padded}${prefix}")
-    endif()
-    unset(version_str)
-    unset(version_padded)
-    unset(version_req_str)
-  endforeach()
-endif()
+include(FairSummary)
```
rbx, line 250: here.

CMakeLists.txt
```diff
@@ -16,13 +16,25 @@ include(FairMQLib)
...
 
 project(FairMQ VERSION ${PROJECT_VERSION} LANGUAGES CXX)
-message(STATUS "${BWhite}${PROJECT_NAME}${CR} ${PROJECT_GIT_VERSION} from ${PROJECT_DATE}")
 
-set(PROJECT_MIN_CXX_STANDARD 17)
+include(CTest)
+include(BundlePackages)
+
+set(PROJECT_FairCMakeModules_VERSION 0.2)
+find_package(FairCMakeModules ${PROJECT_FairCMakeModules_VERSION} QUIET)
```
ChristianTackeGSI, line 24: I assume, you're proposing this, so that the Summary has a `>= 0.2` in it?

According to [the docs](https://cmake.org/cmake/help/latest/command/find_package.html#package-file-interface-variables) config mode packages get a variable named `<PackageName>_FIND_VERSION`. So we *could* handle that in FairCMakeModules?

Maybe like this?

```cmake
if(@PROJECT_NAME@_FIND_VERSION)
  set(_@PROJECT_NAME@_REQUESTED_VERION "${FairCMakeModules_FIND_VERSION}")
endif()
```

and in FairFindPackage2.cmake:
```cmake
if(@PROJECT_NAME@_FOUND)
  list(APPEND PROJECT_PACKAGE_DEPENDENCIES @PROJECT_NAME@)
  if(_@PROJECT_NAME@_REQUESTED_VERION)
    set(PROJECT_@PROJECT_NAME@_VERSION "${_@PROJECT_NAME@_REQUESTED_VERION}")
    unset(_@PROJECT_NAME@_REQUESTED_VERION)
  endif()
endif()
```

dennisklein: > I assume, you're proposing this, so that the Summary has a >= 0.2 in it?

yes

dennisklein: good idea

rbx (approved):

CMakeLists.txt
dennisklein: Thx, I added another commit which refactors the whole CMake into topic-based modules and adds include_guards, so each module should now `include()` everything it directly depends on. But this particular include here was just redundant, it is gone.

rbx:
ubuntu box needs xmllint?

dennisklein:
> ubuntu box needs xmllint?

Yes, and also different flatbuffers: https://cdash.gsi.de/viewBuildError.php?buildid=274331

rbx:
I tested this with the latest O2. I don't see any issues.

rbx (dismissed):

PR #391, opened 2021-09-15
Apply some clang-tidy suggestions

rbx:


dennisklein (commented):
Looks good thx!

In the ctor commit: To me it feels like there are a bit too many `= delete`, could be wrong. Have you double checked that no semantics was changed? Even if we wanted to change the semantics (e.g. it was just coincidental up to now), we should do this in a separate commit from this one!

dennisklein:
Failing test reported in #398. Unrelated to this PR.

rbx:
I did consider what it might affect for each ctor. My thinking there was - if it is unused (not only in code, but in our current intended use), delete it. Might've missed smth of course.

dennisklein (approved):
Thx for adding more context to the ctor commit msg.

> These classes don't need to be copyable/movable:

Many of them should really become movable though 😜 - but that is not the scope of this PR.

PR #399, opened 2021-09-30
Add options to control allocation attempts

rbx:
Add options to control allocation attempts:
- `bad-alloc-max-attempts`: Maximum number of allocation attempts to perform, when memory cannot be allocated, before throwing `fair::mq::MessageBadAlloc`. -1 is infinite. At least one attempt is always done, so `0` = `1`.
- `bad-alloc-attempt-interval`: Interval between allocation attempts in milliseconds (default is 50ms).

Currently only shmem transport honors these settings.

The existing `shm-throw-bad-alloc` setting can be completely replaced with the above two, but it is still left for backwards-compatibility (Can be removed in the future), and behaves the following way:
If `shm-throw-bad-alloc` is explicitly set to false (default is true), shmem transport will ignore the two new options and will not throw. If it is left at default `true`, the new options will be honored and throwing can still be disabled by setting `bad-alloc-max-attempts` to `-1`.


dennisklein:
I will look into the failing test first thing tomorrow.

Issue #384, opened 2021-07-21
Make number of connected peers available to the user on a given `fair::mq::Channel`


dennisklein:

davidrohr:
@dennisklein : Can you give us an ETA when this feature might become available? We want to use that information for the EndOfStream handling, to check if there are still peers connected to a channel.

dennisklein:
> Can you give us an ETA when this feature might become available?

Mid to end of next week seems doable.

davidrohr:
Thx, that would be very much appreciated. Nice to have such a short response time :)

PR #401, opened 2021-10-13
Make number of connected peers available

dennisklein:
* [x] Add a button to the interactive controller that prints it
* [x] More extensive testing needed (only did the bare minimum in the tests so far)
* [x] Have not thought a lot about thread-safety yet
* [x] @ktf Is a blocking API acceptable or do you need anything asynchronous for this?

implements #384

### DIFF
--- a/fairmq/Device.h
+++ b/fairmq/Device.h
@@ -327,6 +327,14 @@ class Device
         throw;
     }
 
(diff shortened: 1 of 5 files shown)


dennisklein:
Improvements will follow later via #402.

PR #405, opened 2021-11-11
Add configurable default snd/rcv timeout

rbx:
- Add default send/receive timeout, configurable via `--channel-config` keys `sndTimeoutMs` and `rcvTimeoutMs`. Only affects default value (which is `-1` - no timeout - by default), when no explicit timeout argument is provided to Send/Receive.


dennisklein (commented):
Would a `std::chrono::milliseconds` instead of `int` be feasible?

rbx:
In principle yes. But since the other transfer timeouts are int i made this int also to be consistent. We can move all of them to chrono in one go. Or if it can be non-breaking I can add the full move as a commit here.

dennisklein:
> Or if it can be non-breaking I can add the full move as a commit here.

Probably only as additional overloads.

rbx:
I added another commit with some more zmq/shmem deduplications.

PR #412, opened 2022-01-12
Fix regression in region events handler

rbx:
- Fix regression in region events handler (double unlock).
- shm: keep also the mng segment around when skipping cleanup.
- expand keep-alive example exe a bit.


rbx:
All tests pass (except the usual sanitizer suspects), but fail to submit results.
Merging.

PR #420, opened 2022-03-03
fix: Data races and deadlocks

dennisklein:
Starting to fix thread sanitizer errors, built with `-DBUILD_TESTING=ON -DCMAKE_BUILD_TYPE=Debug -DENABLE_SANITIZER_THREAD=ON` and then running the test suite. Got it down from 1550 (319 unique) errors to 113 (73 unique). Still too much to post.


dennisklein:
included in #422

PR #435, opened 2022-08-17
New tunable `--shm-metadata-msg-size`

dennisklein:
tests/docs and multipart send coming tomorrow, but should already be enough for a first round of discussion.


fairmq/shmem/Socket.h
```diff
@@ -125,23 +135,83 @@ class Socket final : public fair::mq::Socket
...
+        if (!msgPtr) {
+            return static_cast<int>(TransferCode::error);
+        }
+        assertm(dynamic_cast<shmem::Message*>(msgPtr), "given mq::Message is a shmem::Message");   // NOLINT
+        auto shmMsg = static_cast<shmem::Message*>(msgPtr);   // NOLINT(cppcoreguidelines-pro-type-static-cast-downcast)
+
+        auto metaMsg = MakeMetaMsg(*shmMsg);
+        // according to http://api.zeromq.org/master:zmq-msg-close we need to zmq_msg_close in all cases
+        // except on successfull send to not leak the msg.
+        auto closeMetaMsg = tools::CallOnDestruction([&metaMsg]() { zmq_msg_close(&metaMsg); });
```
dennisklein, line 199: This is ugly. @rbx do you know if I can just always call `zmq_msg_close`?

On a second note, should I perhaps use [`shmem::ZMsg`](https://github.com/FairRootGroup/FairMQ/blob/73fd1b2c2a3c2784368aba2c7805b70c6ed9afdf/fairmq/shmem/Socket.h#L33) here?

fairmq/shmem/Socket.h
```diff
@@ -125,23 +135,83 @@ class Socket final : public fair::mq::Socket
...
+        zmq_msg_t metaMsg;
+        auto rc = 0;
+
+        if (fMetadataMsgSize > size) {
+            // copy the MetaHeader into a larger buffer than needed
+            auto paddedMetaMsg = tools::MakePaddedUnique<MetaHeader>(
+                size, alignof(std::max_align_t), std::pmr::get_default_resource(), shmMsg.fMeta);
+
+            using Deleter = typename decltype(paddedMetaMsg)::deleter_type;
+            auto del = std::make_unique<Deleter>(paddedMetaMsg.get_deleter());
```
dennisklein, line 151: This extra allocation I would like to avoid still, was thinking of copying the deleter into the free buffer space after the `MetaHeader`. Just I am not sure, if this would be UB, if the deleter would deallocate itself "as last action" (and also what about calling the dtor?!).

fairmq/tools/MakeUnique.h
```diff
@@ -0,0 +1,41 @@
...
+// * alignment, and
+// * right padding (allocating a larger heap buffer than strictly needed for the
+//   given T)
+template<typename T, typename... Args, typename Allocator = std::pmr::memory_resource>
+constexpr auto MakePaddedUnique(std::size_t size_hint, /* = sizeof(T) */
+                                std::size_t alignment, /* = alignof(std::max_align_t) */
+                                Allocator* pmr,        /* = std::pmr::get_default_resource() */
+                                Args&&... args)
+{
+    auto const size = (size_hint > sizeof(T)) ? size_hint : sizeof(T);
```
rbx, line 31: Would it be more clear to call this entire option `shm-metadata-padding`?
1. to me it is more clear from the name what it does. With the current name, having default size at 0 is kind of strange, although you do explain the condition in the docs.
2. you don't need the if/else here, just size+padding.

fairmq/shmem/Socket.h
rbx: I understand the docs in the way that you can call zmg_msg_close when *you* no longer need the msg. It will still live on until it is transferred. So always closing is fine.

dennisklein: done

fairmq/tools/MakeUnique.h
dennisklein: I see your point, however, I kept the command line option as is (`--shm-metadata-msg-size`), because from there the user does not know anything about the actual size of the metadata msg, so it would be strange to just tune the padding. But in the C++ interfaces I changed it to padding, because it simplifies the code.

fairmq/shmem/Socket.h
```diff
@@ -125,25 +116,37 @@ class Socket final : public fair::mq::Socket
...
-        int flags = 0;
-        if (timeout == 0) {
-            flags = ZMQ_DONTWAIT;
+        auto const size = std::max(fMetadataMsgSize, sizeof(MetaHeader));
+
+        auto buffer = std::malloc(size);   // owning raw pointer, NOLINT
+        if (!buffer) {
+            LOG(error) << "Could not allocate shmem meta msg buffer, terminating ...";
+            std::terminate();
         }
```
dennisklein, line 128: @rbx Do we want to handle this more gracefully? Returning a Send failure to the user perhaps?

rbx: You could return `TransferCode::error`, but if malloc fails idk if it will help anything.

dennisklein: > but if malloc fails idk if it will help anything.

Was thinking the same, I guess it's fine to terminate here.

fairmq/Socket.h
```diff
@@ -27,7 +27,8 @@ enum class TransferCode : int
     success = 0,
     error = -1,
     timeout = -2,
-    interrupted = -3
+    interrupted = -3,
+    tryagain = -4
```
rbx, line 31: What was this for again?

rbx:
When I run our benchmark (`./fairmq/startMQBenchmark.sh 1000000 0 shmem true`) with this branch, plus:
```diff
SAMPLER+=" --multipart $multipart"
SAMPLER+=" --num-parts $numParts"
SAMPLER+=" --shm-throw-bad-alloc false"
+ SAMPLER+=" --shm-metadata-msg-size 1024"
# SAMPLER+=" --msg-rate 1000"
SAMPLER+=" --max-iterations $maxIterations"
SAMPLER+=" --channel-config name=data,type=pair,method=bind,address=tcp://127.0.0.1:5555"

```
, I immediately get from the sampler:
```
free(): double free detected in tcache 2
```
Multipart mode (`./fairmq/startMQBenchmark.sh 1000000 0 shmem true true 2`) runs fine, but when leaving Running state, sometimes there's an exception:
```
[12:41:36][ERROR] Uncaught exception reached the top of main: shmem: could not create a message of size 1000000, alignment: 16, free memory: 2146483456
```
Probably it makes sense if the message allocation was interrupted by the stop signal. However the benchmark is executed with `--shm-throw-bad-alloc false` so then it should cancel silently?

fairmq/Socket.h
dennisklein: not needed, removed it.

rbx:
> True, however, I tend to say this is not related to this PR - at least I don't see how. Also, I get it with v1.5.0. Seems, we broke it somewhere else?

Oh ok, I didn't try enough times. Can confirm that it also happens on the dev branch.

rbx:
> > , I immediately get from the sampler:
> > ```
> > free(): double free detected in tcache 2
> > ```
> 
> This should be fixed by the [latest push](https://github.com/FairRootGroup/FairMQ/compare/726873cda19d124187a55b805f174f9cd301dd9d..a184677c91a750f608da3c2e2ae9a2125f72778d)

Now the sink fails with:
```
[15:25:36][ERROR] Could not get segment with id '63597': Unable to find key in unordered_map.
terminate called after throwing an instance of 'std::out_of_range'
  what():  _Map_base::at
```
Seems that it's not reading the incoming data correctly, there is no segment with id `63597`

dennisklein:
> > > , I immediately get from the sampler:
> > > ```
> > > free(): double free detected in tcache 2
> > > ```
> > 
> > 
> > This should be fixed by the [latest push](https://github.com/FairRootGroup/FairMQ/compare/726873cda19d124187a55b805f174f9cd301dd9d..a184677c91a750f608da3c2e2ae9a2125f72778d)
> 
> Now the sink fails with:
> 
> ```
> [15:25:36][ERROR] Could not get segment with id '63597': Unable to find key in unordered_map.
> terminate called after throwing an instance of 'std::out_of_range'
>   what():  _Map_base::at
> ```
> 
> Seems that it's not reading the incoming data correctly, there is no segment with id `63597`

Ah yes, forgot to mention. Still looking at that.

knopers8:
Hello, would you have an ETA of this by any chance?

dennisklein:
> Hello, would you have an ETA of this by any chance?

Let's make it highest priority. It is pending for way too long. Let me prepare another update on this PR branch (will need till tomorrow). Then let's check it against with your use case and if successful, push it with a 1.6 release (if all goes well, we can do the release on monday).

knopers8:
Thank you!

dennisklein:
@knopers8 I haven't been able to push a release until now. However, I think, with the latest update, this is close to become mergeable. Also, @rbx will be back soon and let's see what he is thinking. Since I am on a trip from tomorrow till sunday, I can only continue next Monday.

@rbx I intentionally removed some of the [already merged api removal commits in `dev`](https://github.com/FairRootGroup/FairMQ/pull/472) from this PR branch with the intent to force-push this to `dev` once its ready to merge (so, don't use the gh merge button for now). The plan is to release this feature as [v1.6.0](https://github.com/FairRootGroup/FairMQ/milestone/5) to have a minimal diff to v1.5.0, and then make a [v2.0.0](https://github.com/FairRootGroup/FairMQ/milestone/6) release with the deprecated api removals later.

fairmq/plugins/config/Config.cxx
```diff
@@ -72,6 +73,7 @@ Plugin::ProgOptions ConfigPluginProgramOptions()
         ("shm-zero-segment",              po::value<bool          >()->default_value(false),             "Shared memory: zero the shared memory segment memory after initialization (opened or created).")
         ("shm-zero-segment-on-creation",  po::value<bool          >()->default_value(false),             "Shared memory: zero the shared memory segment memory only once when created.")
         ("shm-throw-bad-alloc",           po::value<bool          >()->default_value(true),              "Shared memory: throw fair::mq::MessageBadAlloc if cannot allocate a message (retry if false).")
+        ("shm-metadata-msg-size",         po::value<std::size_t   >()->default_value(0),                 "Shared memory: size of the zmq metadata message (values smaller than minimum are clamped to the minimum). MAY NOT BE IMPLEMENTED ON SOME libc++ PLATFORMS!")
```
rbx, line 76: What is the reason for the libc++ note?

dennisklein: Obsolete, will remove it. Was from an earlier implementation which used the PMR std library, which is not implemented everywhere.

fairmq/zeromq/Common.h
```diff
@@ -87,6 +92,16 @@ inline int HandleErrors(const std::string& id)
...
 
+__FAIRMQ_ALWAYS_INLINE auto ZSend(void* socket, UMsg&& msg, int flags)
+{
+    return zmq_send(socket, msg.Data(), msg.Size(), flags);
+}
+
+__FAIRMQ_ALWAYS_INLINE auto ZSend(void* socket, ZMsg&& msg, int flags)
+{
+    return zmq_msg_send(msg.Msg(), socket, flags);
+}
```
dennisklein, line 103: I can also remove those. They were more complicated earlier, but got simplified to a one-liner, which makes the `ZSend` not that useful any more.

rbx: Great, I wanted to ask about these too!

dennisklein:
@knopers8 Currently, there is no way to tune this per channel, only per device and applies to all sending shmem channels in that device. Is this granular enough?

fairmq/shmem/Socket.h
```diff
@@ -266,35 +291,31 @@ class Socket final : public fair::mq::Socket
 
     int64_t Receive(std::vector<MessagePtr>& msgVec, int timeout = -1) override
     {
-        int flags = 0;
-        if (timeout == 0) {
-            flags = ZMQ_DONTWAIT;
-        }
-        int elapsed = 0;
+        assert(msgVec.empty());
```
rbx, line 294: I think previously it was fine to receive into a non-empty vector. Is it different now? `reserve` would behave not as expected, but in principle it should work, or?

We had a request from Alice in the past (few years back) to allow receiving into a non-empty vector.

dennisklein: I guess, I can just `msgVec.reserve(msgVec.size() + n);` and remove the pre-condition.

fairmq/shmem/Socket.h
```diff
@@ -211,28 +249,21 @@ class Socket final : public fair::mq::Socket
...
-            auto shmMsg = static_cast<shmem::Message*>(msgPtr);   // NOLINT(cppcoreguidelines-pro-type-static-cast-downcast)
-            std::memcpy(metas++, &(shmMsg->fMeta), sizeof(MetaHeader));
         }
 
+        auto metaMsg = MakeMetaMsg(msgVec);
+
         while (true) {
             int64_t totalSize = 0;
-            int nbytes = zmq_msg_send(zmqMsg.Msg(), fSocket, flags);
+            int nbytes = zmq_send(fSocket, metaMsg.Data(), metaMsg.Size(), flags);
```
dennisklein, line 264: @rbx Hm, this makes an additional copy, right? So, I should use a `ZMsg` here.

fairmq/shmem/Socket.h
```diff
@@ -126,6 +115,24 @@ class Socket final : public fair::mq::Socket
         return zmq::Connect(fSocket, address, fId);
     }
 
+  private:
+    __FAIRMQ_ALWAYS_INLINE auto MakeMetaMsg(shmem::Message const& shmMsg) noexcept
+    {
+        if (fMetadataMsgSize > sizeof(MetaHeader)) {
+            zmq::UMsg metaMsg(fMetadataMsgSize);
```
dennisklein, line 122: This should probably also be a `ZMsg` to avoid a copy.

fairmq/shmem/Socket.h
rbx: True

fairmq/shmem/Socket.h
rbx: Yeah. Also in the vector case?

fairmq/shmem/Socket.h
dennisklein: Switched all cases to `ZMsg` which obsoleted `UMsg` so I removed it again. Will need another round of testing, also need to cover the new code paths in some of our unit tests.

dennisklein:
@rbx thx for the test coverage. I shortened the code even more and added the changes from what we discussed earlier today.

rbx:
Looks good!

PR #436, opened 2022-08-18
fix(shm): race/deadlock in region locks

rbx:
Fixes a race/deadlock in GetRegion/GetRegionInfo.


dennisklein (changes requested):
hm, your commit changed the googletest submodule. If that was intentional, please commit it as a separate commit. thx!

PR #438, opened 2022-08-23
New example with a custom controller

dennisklein:
New example demonstrating a simple custom controller (statically compiled into the executable) which executes a simple static plan through the state machine (similar to the built-in `--control static`). Builtin signal handling is disabled.


rbx:
Why is there the extra commit from the shmem fix? It is already in the dev.

dennisklein:
> Why is there the extra commit from the shmem fix? It is already in the dev.

Because GH renders it strangely. There is a merge commit, so that dev can see the latest tag. But as you can see in the files tab, those commits are not adding changes.

rbx (approved):

PR #440, opened 2022-08-26
Reproducer for control misbehavior

ktf:
This is a reproducer for my issue in DPL with the interactive control. You can run it with:

./examples/1-1/fairmq-ex-1-1-sampler --control interactive

and it will exhibit the same unexpected behaviour:

```
[09:37:01][STATE] INITIALIZING TASK ---> READY
[09:37:01][STATE] READY ---> RUNNING
[09:37:01][INFO] Switching toSTOP

[09:37:01][INFO] fair::mq::Device running...
[09:37:01][STATE] RUNNING ---> READY
[09:37:01][INFO] Switching toRESET TASK

[09:37:01][STATE] READY ---> RESETTING TASK
[09:37:01][INFO] Switching toRESET DEVICE

[09:37:01][STATE] No transition from state RESETTING_TASK on transition RESET_DEVICE
[09:37:01][STATE] RESETTING TASK ---> DEVICE READY
[09:37:01][INFO] Switching toEND

[09:37:01][STATE] No transition from state DEVICE_READY on transition END
```

### DIFF
--- a/examples/1-1/sampler.cxx
+++ b/examples/1-1/sampler.cxx
@@ -8,8 +8,9 @@
 
 #include <fairmq/Device.h>
 #include <fairmq/runDevice.h>
-
+#include <functional>
 #include <string>
+#include <vector>
 
 namespace bpo = boost::program_options;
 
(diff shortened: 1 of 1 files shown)


ktf:
Again I think the issue is the the automatic transitions do not take into account the original state of a given transition.

examples/1-1/sampler.cxx
```diff
@@ -8,15 +8,31 @@
...
+        std::function<void(const fair::mq::State)> stateWatcher =
+            [this](const fair::mq::State state) -> void {
+            if (state == fair::mq::State::Running) {
+                isRunning = true;
+            }
+            if (isRunning) {
+                LOG(info) << "Switching to " << transitions.back() << std::endl;
+                assert(transitions.empty() == false);
+                ChangeState(transitions.back());
+                transitions.pop_back();
```
dennisklein, line 32: Most simple fix I can think of:
```suggestion
                if (ChangeState(transitions.back())) {
                    transitions.pop_back();
                }
```

dennisklein:
We discussed the topic at length in a private mm chat last week, for agreed upon improvements see #441, #442, #443.

PR #453, opened 2022-10-04
Shm: error on duplicate region IDs

rbx:
- Error on duplicate unmanaged region IDs.
- Extend debug output slightly.
- Add a test for externally (outside of a session) created shmem.

### DIFF
--- a/examples/region/sampler.cxx
+++ b/examples/region/sampler.cxx
@@ -39,12 +39,9 @@ struct Sampler : fair::mq::Device
         if (fExternalRegion) {
             regionCfg.id = 1;
             regionCfg.removeOnDestruction = false;
-            regionCfg.lock = false; // mlock region after creation
-            regionCfg.lock = false; // mlock region after creation
-        } else {
-            regionCfg.lock = true; // mlock region after creation
-            regionCfg.zero = true; // zero region content after creation
         }
+        regionCfg.lock = !fExternalRegion; // mlock region after creation
+        regionCfg.zero = !fExternalRegion; // zero region content after creation
         fRegion = fair::mq::UnmanagedRegionPtr(NewUnmanagedRegionFor(
             "data", // region is created using the transport of this channel...
             0,      // ... and this sub-channel
--- a/fairmq/shmem/Manager.h
+++ b/fairmq/shmem/Manager.h
@@ -396,7 +396,7 @@ class Manager
                 const uint16_t id = cfg.id.value();
 
                 std::lock_guard<std::mutex> lock(fLocalRegionsMtx);
-                auto& region = fRegions[id] = std::make_unique<UnmanagedRegion>(fShmId, size, false, cfg);
+                auto& region = fRegions[id] = std::make_unique<UnmanagedRegion>(fShmId, size, true, cfg);
                 // LOG(debug) << "Created region with id '" << id << "', path: '" << cfg.path << "', flags: '" << cfg.creationFlags << "'";
 
                 // start ack receiver only if a callback has been provided.
@@ -463,7 +463,7 @@ class Manager
                 }
                 // LOG(debug) << "Located remote region with id '" << id << "', path: '" << cfg.path << "', flags: '" << cfg.creationFlags << "'";
 
-                auto r = fRegions.emplace(id, std::make_unique<UnmanagedRegion>(fShmId, 0, true, std::move(cfg)));
+                auto r = fRegions.emplace(id, std::make_unique<UnmanagedRegion>(fShmId, 0, false, std::move(cfg)));
                 r.first->second->InitializeQueues();
                 r.first->second->StartAckSender();
                 return r.first->second.get();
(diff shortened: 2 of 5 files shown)


fairmq/shmem/Manager.h
```diff
@@ -396,7 +396,23 @@ class Manager
...
-                auto& region = fRegions[id] = std::make_unique<UnmanagedRegion>(fShmId, size, true, cfg);
+
+                UnmanagedRegion* region = nullptr;
+
+                auto it = fRegions.find(id);
+                if (it != fRegions.end()) {
+                    region = it->second.get();
+                    if (region->fControlling) {
+                        LOG(error) << "Unamanged Region with id " << cfg.id.value() << " already exists. Only unique IDs per session are allowed.";
+                        throw TransportError(tools::ToString("Unamanged Region with id ", cfg.id.value(), " already exists. Only unique IDs per session are allowed."));
```
dennisklein, line 407:
```suggestion
                        throw TransportError(tools::ToString("Unmanaged Region with id ", cfg.id.value(), " already exists. Only unique IDs per session are allowed."));
```

fairmq/shmem/UnmanagedRegion.h
```diff
@@ -263,9 +270,13 @@ struct UnmanagedRegion
...
-        bool newShmRegionCreated = shmRegions->emplace(cfg.id.value(), RegionInfo(cfg.path.c_str(), cfg.creationFlags, cfg.userFlags, cfg.size, alloc)).second;
-        if (newShmRegionCreated) {
+        auto it = shmRegions->find(cfg.id.value());
+        if (it == shmRegions->end()) {
+            shmRegions->emplace(cfg.id.value(), RegionInfo(cfg.path.c_str(), cfg.creationFlags, cfg.userFlags, cfg.size, alloc)).second;
             (eventCounter->fCount)++;
+        } else {
+            LOG(error) << "Unamanged Region with id " << cfg.id.value() << " has already been registered. Only unique IDs per session are allowed.";
+            throw TransportError(tools::ToString("Unamanged Region with id ", cfg.id.value(), " has already been registered. Only unique IDs per session are allowed."));
         }
```
dennisklein, line 280:
```suggestion
        if (it != shmRegions->end()) {
            LOG(error) << "Unmanaged Region with id " << cfg.id.value() << " has already been registered. Only unique IDs per session are allowed.";
            throw TransportError(tools::ToString("Unamanged Region with id ", cfg.id.value(), " has already been registered. Only unique IDs per session are allowed."));
        }
        
        shmRegions->emplace(cfg.id.value(), RegionInfo(cfg.path.c_str(), cfg.creationFlags, cfg.userFlags, cfg.size, alloc)).second;
        (eventCounter->fCount)++;
```

Note: Even if you don't like this, there is also the "Unamanged" typo here.

fairmq/shmem/Manager.h
```diff
@@ -396,7 +396,23 @@ class Manager
...
+                    } else {
+                        LOG(debug) << "Unmanaged region (view) already present, promoting to controller";
+                        region->BecomeController(cfg);
+                    }
```
dennisklein, line 411:
```suggestion
                    }
                    
                    LOG(debug) << "Unmanaged region (view) already present, promoting to controller";
                    region->BecomeController(cfg);
```

dennisklein (changes requested):
What do you think about this:

```cpp
struct TransportError : ::std::runtime_error
{
    template<typename... Args>
    explicit TransportError(Args&&... args)
        : ::std::runtime_error::runtime_error(tools::ToString(std::forward<Args>(args)...))
    {
        LOG(error) << what();
    }
};
```

to deduplicate the error case code or will that mess with the logger file/linenr macro?

Somehow logging and throwing feels wrong, shouldn't we just log when catching this exception?

examples/region/sampler.cxx
```diff
@@ -40,7 +40,7 @@ struct Sampler : fair::mq::Device
             regionCfg.id = 1;
             regionCfg.removeOnDestruction = false;
             regionCfg.lock = false; // mlock region after creation
-            regionCfg.lock = false; // mlock region after creation
+            regionCfg.zero = false; // zero region content after creation
         } else {
             regionCfg.lock = true; // mlock region after creation
             regionCfg.zero = true; // zero region content after creation
```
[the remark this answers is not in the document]
rbx: applied

fairmq/shmem/UnmanagedRegion.h
rbx: applied

fairmq/shmem/Manager.h
rbx: applied

rbx:
> What do you think about this:
> 
> ```c++
> struct TransportError : ::std::runtime_error
> {
>     template<typename... Args>
>     explicit TransportError(Args&&... args)
>         : ::std::runtime_error::runtime_error(tools::ToString(std::forward<Args>(args)...))
>     {
>         LOG(error) << what();
>     }
> };
> ```
> 
> to deduplicate the error case code or will that mess with the logger file/linenr macro?
> 
> Somehow logging and throwing feels wrong, shouldn't we just log when catching this exception?

Log & throw doesn't really bother me much - for unrecoverable exceptions. At least this way it is sure that the error is printed if the exception is uncaught and/or terminates a thread.

dennisklein (approved):

PR #456, opened 2022-11-17
Warn when going to try allocate a shmem message more times

knopers8:
When FairMQ is configured to try allocate a shmem message indefinitely, we might end up in a situation when a device is stuck without any information why it could be so. This log would help us understand why a device is stuck.

### DIFF
--- a/fairmq/shmem/Manager.h
+++ b/fairmq/shmem/Manager.h
@@ -753,6 +753,9 @@ class Manager
                 if (fBadAllocMaxAttempts >= 0 && ++numAttempts >= fBadAllocMaxAttempts) {
                     throw MessageBadAlloc(tools::ToString("shmem: could not create a message of size ", size, ", alignment: ", (alignment != 0) ? std::to_string(alignment) : "default", ", free memory: ", boost::apply_visitor(SegmentFreeMemory(), fSegments.at(fSegmentId))));
                 }
+                if (numAttempts == 1 && fBadAllocMaxAttempts > 1) {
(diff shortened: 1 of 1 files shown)


rbx:
As you can see three lines above, I already had a warning messages (although less detailed one, and on every iteration) for such cases.
But there can be scenarios where it is valid to just wait and idle until eventually memory is available (or not). And there the warning is not useful.

So for your case I would rather let it throw the exception, catch it and add the warning there.

rbx:
Would the debug severity instead of warning be sufficient for you?

knopers8:
>  But there can be scenarios where it is valid to just wait and idle until eventually memory is available (or not). And there the warning is not useful.

Yes, this is why we use it this way :) And by adding just one log, not for each try, we should avoid the spam.

> Would the debug severity instead of warning be sufficient for you?

AFAICT, we disable debug messages from FairMQ in the production setup, so that would not help much unfortunately. I am not sure if it would make sense to enable FairMQ debug messages there just to be able to see this one.

If a warning is a "no" to you, then I will just go for the catch-throw approach. Not the cleanest to me, but should probably do.

rbx:
> AFAICT, we disable debug messages from FairMQ in the production setup, so that would not help much unfortunately. I am not sure if it would make sense to enable FairMQ debug messages there just to be able to see this one.

Mmm, right, there was that... Ok, let's leave it at the warning severity, at least if it is once per call it is not a spammy one (like the one I had there before).

rbx (approved):

knopers8:
Thanks! And sorry I did not realize you are using `dev` :)

Issue #428, opened 2022-05-03
Rework transport interrupt/resume periods


dennisklein:
The internal transport interrupt/resume feature was introduced and built around the assumption that messages are only sent/received in the `RUNNING` state. However, nowadays there are use case, when message transfers are also used outside of the `RUNNING` state.

The additional internal 100ms timeout - when a transport is interrupted - is therefor active in those outside-of-RUNNING-msg-transfers while not really needed. A rework shall limit the active interrupted state of the transports to `NewStatePending() == true` conditions only and apply this consistently for all states.

PR #460, opened 2023-02-22
Resume/Interrupt transports consistently

rbx:
- Resume transports before state callbacks & handlers.
 - Interrupt transports on new transitions.

Solves #428.

### DIFF
--- a/fairmq/Device.cxx
+++ b/fairmq/Device.cxx
@@ -650,6 +650,8 @@ bool Device::HandleMultipartInput(const string& chName, const InputMultipartCall
 
 shared_ptr<TransportFactory> Device::AddTransport(mq::Transport transport)
 {
+    lock_guard<mutex> lock(fTransportMtx);
+
     if (transport == mq::Transport::DEFAULT) {
         transport = fDefaultTransportType;
     }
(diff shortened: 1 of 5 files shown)


rbx:
The failures happen because there is a race:
1. `Device::ResetWrapper()` is clearning the transports container.
2. `Control::RunShutdownSequence()` of a plugin is calling `ChangeDeviceState(DeviceStateTransition::Auto);`, which leads `Device::InterruptTransports();` via `SubscribeToNewTransition(`.
So, one is accessing transports while the other is deleting it.

For now just some notes:

- ShutdownSequence doesn't really need these:
```
case DeviceState::Binding:
case DeviceState::Connecting:
case DeviceState::InitializingTask:
case DeviceState::ResettingTask:
case DeviceState::ResettingDevice:
    ChangeDeviceState(DeviceStateTransition::Auto);
```
, since those auto-transition anyway.
- `InterruptTransports` is not needed on the auto-transition from ResettingDevice, but the auto transitions should be disambiguated.

dennisklein:
> So, one is accessing transports while the other is deleting it.

I guess we should synchronize those, shouldn't we?

rbx:
> > So, one is accessing transports while the other is deleting it.
> 
> I guess we should synchronize those, shouldn't we?

Yeah.
Also there is another race condition:
2023-02-22 17_24_18-Resume_Interrupt transports consistently by rbx · Pull Request #460 · FairRootGr
2023-02-22 17_24_37-Resume_Interrupt transports consistently by rbx · Pull Request #460 · FairRootGr

dennisklein (approved):
Looks good, except I don't really understand 0c5d9a73b77204bb7fc4e1a882ede274051d701a

Issue #459, opened 2023-02-21
Make --shm-monitor default to false


rbx:
It cannot be relied upon if control system decided to kill all tasks (including the monitor).
And occasionally it can also remove shmem when it should not (e.g. due to delayed heartbeats from devices).
Examples & tests will need to be adapted to run cleanup on success/failure.

PR #461, opened 2023-02-24
Turn autolaunched fairmq-shmmonitor off

rbx:
- Turn `--shm-monitor` to `false` by default. Turn it on for tests.
- Check exit codes in the example tests.
- Update shmem docs.


dennisklein (commented):
Please add a commit msg body `resolves #459` to 9c484a521ca96ff73266bf4f549613beddef0d09

examples/multipart/sink.cxx
```diff
@@ -24,6 +24,8 @@ struct Sink : fair::mq::Device
     {
         LOG(info) << "Received message with " << parts.Size() << " parts";
 
+        assert(parts.Size() == 7);
```
dennisklein, line 27: hm, this will be compiled away if `NDEBUG` is set, won't it?

rbx: true, I'll change it.

dennisklein (approved):

PR #469, opened 2023-03-02
Migrate to `std::filesystem` consistently

dennisklein:


dennisklein:
```
[----------] 1 test from PropertyHelper
[ RUN      ] PropertyHelper.ConvertPropertyToString
/Users/alfaci/alfaci/workspace/FairRootGroup_FairMQ_PR-469/test/properties/_properties.cxx:215: Failure
Expected equality of these values:
  PropertyHelper::ConvertPropertyToString(Property(fs::path("C:\\Windows")))
    Which is: "\"C:\\\\Windows\""
  "\"C:\\Windows\""
/Users/alfaci/alfaci/workspace/FairRootGroup_FairMQ_PR-469/test/properties/_properties.cxx:234: Failure
Expected equality of these values:
  PropertyHelper::ConvertPropertyToString(Property(vector<fs::path>({ fs::path("C:\\Windows"), fs::path("C:\\Windows\\System32") })))
    Which is: "\"C:\\\\Windows\", \"C:\\\\Windows\\\\System32\""
  "\"C:\\Windows\", \"C:\\Windows\\System32\""
[  FAILED  ] PropertyHelper.ConvertPropertyToString (0 ms)
[----------] 1 test from PropertyHelper (0 ms total)
```

@rbx You can switch between `boost::filesystem` and `std::filesystem` now via `-DFAIRMQ_HAS_STD_FILESYSTEM=0|1`. (by default it is 1 if gcc >=9 or clang >= 9)

If I switch to boost, the test succeeds. So, they have the same API, but it produces different results... any idea? looks like escaping is somehow applied twice.

rbx:
The difference seems to be how any_cast works with boost path and std path:
https://godbolt.org/z/1YK8f4djW

fairmq/Properties.cxx
```diff
@@ -8,7 +8,15 @@
...
 #include <fairmq/Properties.h>
 
+#if FAIRMQ_HAS_STD_FILESYSTEM
+#include <filesystem>
+namespace fs = std::filesystem;
+#else
+#define BOOST_FILESYSTEM_VERSION 3
+#define BOOST_FILESYSTEM_NO_DEPRECATED
 #include <boost/filesystem.hpp>
+namespace fs = ::boost::filesystem;
```
rbx, line 18: I wonder how breaking this is. 
If some use has a program option using boost::filesystem::path.

Couldn't find any users by searching atm.

dennisklein: It will be breaking in general, but user can set `-DFAIRMQ_HAS_STD_FILESYSTEM=0` to switch back to boost and adapt at her own schedule.

dennisklein: And the `fs` alias should not break anything.

dennisklein:
> The difference seems to be how any_cast works with boost path and std path: https://godbolt.org/z/1YK8f4djW

Hm, interesting. Well, I just adapted to test cases to expect different output for std::filesystem. Not sure, what else to do.

rbx:
https://en.cppreference.com/w/cpp/filesystem/path/operator_ltltgtgt
Here it says:
> [std::quoted](http://en.cppreference.com/w/cpp/io/manip/quoted) is used so that spaces do not cause truncation when later read by stream input operator.

So I guess boost does not do the quoted part.

> Hm, interesting. Well, I just adapted to test cases to expect different output for std::filesystem. Not sure, what else to do.

Yeah, let's leave it at that.

PR #479, opened 2023-08-04
codemeta: Add GSI as "maintainer"

ChristianTackeGSI:
GSI is providing resources for maintaining FairMQ. So let's document this in codemeta.json.

- [ ] https://github.com/codemeta/codemeta/pull/250#issuecomment-1660417541
  Codemeta < 4.0 does not allow anything but `Person`s in the `maintainer` field. I personally think, we should just ignore this.

---

Checklist:

* [X] Rebased against `dev` branch
* [X] My name is in the resp. CONTRIBUTORS/AUTHORS file
* [X] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)

### DIFF
--- a/codemeta.json
+++ b/codemeta.json
@@ -9,6 +9,13 @@
   "codeRepository": "https://github.com/FairRootGroup/FairMQ/",
   "issueTracker": "https://github.com/FairRootGroup/FairMQ/issues",
   "identifier": "https://doi.org/10.5281/zenodo.1689985",
+  "maintainer": [
+    {
+      "@type": "ResearchOrganisation",
+      "@id": "https://ror.org/02k8cbn47",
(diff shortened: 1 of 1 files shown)


ChristianTackeGSI:
I had thought, that this would conflict with #478. But that's not true. So the only problematic point is Codemeta 4.0 being needed officially (which is not even released), and OSSR sticking at 2.0. But as I already mentioned above, I don't think, we should care too much.

ChristianTackeGSI:
Having looked more at it, the [`sponsor`](https://schema.org/sponsor) field could also be an option. Pro: It already allows "Organization"s today.

To be honest, I prefer the `maintainer` field for FairMQ.

dennisklein (approved):

PR #481, opened 2023-09-05
Allow sorting StateChange callbacks

ktf:


ktf:
Is the thread sanitizer known to be broken?

dennisklein:
> Is the thread sanitizer known to be broken?

yes, it was never green.

ktf:
Can we have a release with this in? In principle it only affects the case in which you name your callback using a numeric prefix.

rbx:
It is now available with v1.8.0. It also contains your previous contribution.

ktf:
Thanks.

PR #483, opened 2023-09-11
Generate `codemeta.json`

dennisklein:


meta_update.py
```diff
@@ -150,7 +151,10 @@ def main():
         if args.newversion is not None:
             manipulator.version(args.newversion)
         manipulator.update_authors()
-        manipulator.save()
+        filename = None
+        if args.outdir is not None:
+            filename = f'{args.outdir}/{manipulator.default_filename}'
```
ChristianTackeGSI, line 156: I would prefer a `pathlib.Path` based approach for this. Yes, not needed for us. But still nicer?

CMakeLists.txt
```diff
@@ -72,6 +72,11 @@ endif()
 if(BUILD_TIDY_TOOL)
   add_subdirectory(fairmq/tidy)
 endif()
+
+add_custom_target(codemeta
+  python3 ${CMAKE_SOURCE_DIR}/meta_update.py --outdir ${CMAKE_BINARY_DIR} --set-version ${PROJECT_VERSION}
```
ChristianTackeGSI, line 77: I would expect a `find_package(2)(Python … something interpreter)` for this, right?

.github/workflows/codemeta.yaml
```diff
@@ -0,0 +1,46 @@
...
+    needs: generate
+    runs-on: ubuntu-latest
+    container:
+      image: gitlab-registry.in2p3.fr/escape2020/wp3/eossr:v1.0
+    steps:
+      - uses: actions/download-artifact@v3
+        with:
+          name: codemeta.json
+      - name: validate codemeta
+        run: eossr-metadata-validator codemeta.json
```
ChristianTackeGSI, line 46: I can somewhat understand, why you moved it here…

I am missing the interdependency between the two jobs?

If there is none: Maybe leave validation part in its own workflow file?

.github/workflows/codemeta.yaml
```diff
@@ -44,3 +44,14 @@ jobs:
           name: codemeta.json
       - name: validate codemeta
         run: eossr-metadata-validator codemeta.json
+  publish:
+    needs: validate
```
ChristianTackeGSI, line 48: Okay, now you have the dep. Maybe do the moving of the validate step in this commit instead?

.github/workflows/codemeta.yaml
```diff
@@ -0,0 +1,46 @@
+name: codemeta
+
+on:
+  push:
+    paths:
+      - codemeta.json
+      - .github/workflows/codemeta.yaml
+    tags:
+      - 'v*'
```
ChristianTackeGSI, line 9: This changes the semantics (AND) compared to the "old" validate workflow. Please comment on this in the commit text body. :-)

ChristianTackeGSI (commented):
Okay. Some comments.

If you want to fix some of them later and consider this a "I need testing for tags" thing, then that's fine.

CMakeLists.txt
dennisklein: Ya, was considering it, but there was so much noise with CMake's Python detection in the past that I am afraid it will introduce more maintenance work than hardcoding it - this target really does not need to be super platform independent.

.github/workflows/codemeta.yaml
dennisklein: see lines 37 and lines 42-44

dennisklein: it validates now the generated `codemeta.json`

.github/workflows/codemeta.yaml
dennisklein: thx, ye, will revisit this. thought it was OR logic.

CMakeLists.txt
ChristianTackeGSI: Okay, got that! Please add a comment above it to that fact. Something like "meant for release management, not needed to be as portable".

Have you looked into `add_custom_command(OUTPUT … MAIN_DEPENDENCY …)` or the `SOURCES` part of `add_custom_target`?

.github/workflows/codemeta.yaml
```diff
@@ -0,0 +1,57 @@
...
+        with:
+          fetch-depth: 150
+          fetch-tags: true
+      - run: "git config --system --add safe.directory $GITHUB_WORKSPACE"
+      - name: configure
+        run: "cmake -G Ninja -S $GITHUB_WORKSPACE -B build"
+      - name: generate codemeta.json
+        run: cmake --build build --target codemeta
+      - name: print result
+        run: cat build/codemeta.json
```
ChristianTackeGSI, line 31: Maybe?
```suggestion
        run: diff codemeta.json build/codemeta.json
```

PR #484, opened 2023-09-27
Shm: Introduce first support for pools

rbx:
This uses `boost::interprocess::adaptive_pool` to store reference counts for unmanaged region buffers when these are "softly" copied by FairMQ.

One additional small managed segment per unmanaged region is created to store the reference count pool. Default segment size is 10,000,000 bytes, which can be changed through regionConfig in code. RefCount object size is 2 bytes, so this should be good for up to 5 million counts.

This replaces the old functionality, which would store the ref counts as separate message objects in the main managed segment. Thus, this should reduce fragmentation of the main segment.

I also add an example that uses this, but tests are also already in place, which tested the old implementation.


fairmq/shmem/Common.h
```diff
@@ -237,17 +256,6 @@ struct RegionCounter
...
-{
-    size_t fSize;
-    size_t fHint;
-    boost::interprocess::managed_shared_memory::handle_t fHandle;
-    mutable boost::interprocess::managed_shared_memory::handle_t fShared;
-    uint16_t fRegionId;
-    mutable uint16_t fSegmentId;
-    bool fManaged;
-};
-
```
dennisklein, line 156: This should go into the previous commit.

dennisklein (commented):
I guess you can squash the [third](https://github.com/FairRootGroup/FairMQ/pull/484/commits/356948f7e244ba84592119ab06e9a22275db20a0) and [last](https://github.com/FairRootGroup/FairMQ/pull/484/commits/6c54e6ec9ed917b93f440e328a3a88ed982ac834) commits?

fairmq/shmem/Manager.h
```diff
@@ -845,7 +778,7 @@ class Manager
     uint64_t fShmId64;
     std::string fShmId;
     uint16_t fSegmentId;
-    std::unordered_map<uint16_t, boost::variant<RBTreeBestFitSegment, SimpleSeqFitSegment>> fSegments; // TODO: refactor to use Segment class
+    std::unordered_map<uint16_t, std::variant<RBTreeBestFitSegment, SimpleSeqFitSegment>> fSegments; // TODO: refactor to use Segment class
```
dennisklein, line 781: > `// TODO: refactor to use Segment class`

:+1:

fairmq/shmem/UnmanagedRegion.h
```diff
@@ -61,6 +61,7 @@ struct UnmanagedRegion
         , fStopAcks(false)
         , fName("fmq_" + shmId + "_rg_" + std::to_string(cfg.id.value()))
         , fQueueName("fmq_" + shmId + "_rgq_" + std::to_string(cfg.id.value()))
+        , fRefCountSegmentName("fmq_" + shmId + "_rrc_" + std::to_string(cfg.id.value()))
```
dennisklein, line 64:
```suggestion
        , fName(MakeSegmentName(shmId, "rg", cfg.id.value()))
        , fQueueName(MakeSegmentName(shmId, "rgq", cfg.id.value()))
        , fRefCountSegmentName(MakeSegmentName(shmId, "rrc", cfg.id.value()))
```

```cpp
struct UnmanagedRegion {
  // ...
  private:
    static std::string MakeSegmentName(const std::string& shmId, std::string_view segment, int regionId) {
        return tools::ToString("fmq_", shmId, "_", segment, "_", regionId);
    }
};
```

or similar.

fairmq/shmem/UnmanagedRegion.h
```diff
@@ -186,6 +187,24 @@ struct UnmanagedRegion
...
+    void RemoveRefCount(RefCount* refCount)
+    {
+        refCount->~RefCount();
+        fRefCountPool->deallocate(refCount, 1);
+    }
+
+    uint16_t IncrementRefCount(RefCount* refCount) { return refCount->count.fetch_add(1); }
+    uint16_t DecrementRefCount(RefCount* refCount) { return refCount->count.fetch_sub(1); }
+
+    uint16_t GetRefCount(RefCount* refCount) { return refCount->count.load(); }
```
dennisklein, line 206: Hmm, those look like they should member functions of `RefCount` itself?

fairmq/shmem/UnmanagedRegion.h
```diff
@@ -294,6 +321,30 @@ struct UnmanagedRegion
...
+        }
+    }
+
+    RefCount* GetRefCountAddressFromHandle(const boost::interprocess::managed_shared_memory::handle_t handle)
+    {
+        if (fRefCountPool) {
+            return reinterpret_cast<RefCount*>(fRefCountSegment->get_address_from_handle(handle));
+        } else {
+            return nullptr;
+        }
```
dennisklein, line 340:
```suggestion
        if (fRefCountPool) {
            return reinterpret_cast<RefCount*>(fRefCountSegment->get_address_from_handle(handle));
        }
        return nullptr;
```

see [readability-else-after-return](https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html)

fairmq/shmem/Manager.h
dennisklein, line 1: While I was browsing this [code snippet](https://github.com/rbx/FairMQ/blob/6c54e6ec9ed917b93f440e328a3a88ed982ac834/fairmq/shmem/Manager.h#L493-L519) I had a hard time following all the indentations. What do you think about writing it like this instead (basically inverting two of the `if`-clauses and using `continue` to avoid having too many indentations):

```diff
--- a/fairmq/shmem/Manager.h
+++ b/fairmq/shmem/Manager.h
@@ -492,30 +492,33 @@ class Manager

         // do another iteration outside of shm lock, to fill ptr+size of unmanaged regions
         for (auto& info : result) {
-            if (!info.managed && info.event == RegionEvent::created) {
-                auto cfgIt = regionCfgs.find(info.id);
-                if (cfgIt != regionCfgs.end()) {
-                    UnmanagedRegion* region = nullptr;
-                    std::lock_guard<std::mutex> lock(fLocalRegionsMtx);
-                    auto it = fRegions.find(info.id);
-                    if (it != fRegions.end()) {
-                        region = it->second.get();
-                    } else {
-                        const uint64_t rcSegmentSize = cfgIt->second.rcSegmentSize;
-                        auto r = fRegions.emplace(cfgIt->first, std::make_unique<UnmanagedRegion>(fShmId, 0, false, cfgIt->second));
-                        region = r.first->second.get();
-                        region->InitializeRefCountSegment(rcSegmentSize);
-                        region->InitializeQueues();
-                        region->StartAckSender();
-                    }
+            if (info.managed || info.event != RegionEvent::created) {
+                continue;
+            }

-                    info.ptr = region->GetData();
-                    info.size = region->GetSize();
-                } else {
-                    info.ptr = nullptr;
-                    info.size = 0;
-                }
+            auto cfgIt = regionCfgs.find(info.id);
+            if (cfgIt == regionCfgs.end()) {
+                info.ptr = nullptr;
+                info.size = 0;
+                continue;
+            }
+
+            UnmanagedRegion* region = nullptr;
+            std::lock_guard<std::mutex> lock(fLocalRegionsMtx);
+            auto it = fRegions.find(info.id);
+            if (it != fRegions.end()) {
+                region = it->second.get();
+            } else {
+                const uint64_t rcSegmentSize = cfgIt->second.rcSegmentSize;
+                auto r = fRegions.emplace(cfgIt->first, std::make_unique<UnmanagedRegion>(fShmId, 0, false, cfgIt->second));
+                region = r.first->second.get();
+                region->InitializeRefCountSegment(rcSegmentSize);
+                region->InitializeQueues();
+                region->StartAckSender();
             }
+
+            info.ptr = region->GetData();
+            info.size = region->GetSize();
         }

         return result;
```

dennisklein, line 1: Here would be another chance to apply [readability-else-after-return](https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html):

```diff
--- a/fairmq/shmem/Manager.h
+++ b/fairmq/shmem/Manager.h
@@ -395,33 +395,33 @@ class Manager
         auto it = fRegions.find(id);
         if (it != fRegions.end()) {
             return it->second.get();
-        } else {
-            try {
-                RegionConfig cfg;
-                const uint64_t rcSegmentSize = cfg.rcSegmentSize;
-                // get region info
-                {
-                    boost::interprocess::scoped_lock<boost::interprocess::interprocess_mutex> shmLock(*fShmMtx);
-                    RegionInfo regionInfo = fShmRegions->at(id);
-                    cfg.id = id;
-                    cfg.creationFlags = regionInfo.fCreationFlags;
-                    cfg.path = regionInfo.fPath.c_str();
-                }
-                // LOG(debug) << "Located remote region with id '" << id << "', path: '" << cfg.path << "', flags: '" << cfg.creationFlags << "'";
+        }

-                auto r = fRegions.emplace(id, std::make_unique<UnmanagedRegion>(fShmId, 0, false, std::move(cfg)));
-                r.first->second->InitializeRefCountSegment(rcSegmentSize);
-                r.first->second->InitializeQueues();
-                r.first->second->StartAckSender();
-                return r.first->second.get();
-            } catch (std::out_of_range& oor) {
-                LOG(error) << "Could not get remote region with id '" << id << "'. Does the region creator run with the same session id?";
-                LOG(error) << oor.what();
-                return nullptr;
-            } catch (boost::interprocess::interprocess_exception& e) {
-                LOG(error) << "Could not get remote region for id '" << id << "': " << e.what();
-                return nullptr;
+        try {
+            RegionConfig cfg;
+            const uint64_t rcSegmentSize = cfg.rcSegmentSize;
+            // get region info
+            {
+                boost::interprocess::scoped_lock<boost::interprocess::interprocess_mutex> shmLock(*fShmMtx);
+                RegionInfo regionInfo = fShmRegions->at(id);
+                cfg.id = id;
+                cfg.creationFlags = regionInfo.fCreationFlags;
+                cfg.path = regionInfo.fPath.c_str();
             }
+            // LOG(debug) << "Located remote region with id '" << id << "', path: '" << cfg.path << "', flags: '" << cfg.creationFlags << "'";
+
+            auto r = fRegions.emplace(id, std::make_unique<UnmanagedRegion>(fShmId, 0, false, std::move(cfg)));
+            r.first->second->InitializeRefCountSegment(rcSegmentSize);
+            r.first->second->InitializeQueues();
+            r.first->second->StartAckSender();
+            return r.first->second.get();
+        } catch (std::out_of_range& oor) {
+            LOG(error) << "Could not get remote region with id '" << id << "'. Does the region creator run with the same session id?";
+            LOG(error) << oor.what();
+            return nullptr;
+        } catch (boost::interprocess::interprocess_exception& e) {
+            LOG(error) << "Could not get remote region for id '" << id << "': " << e.what();
+            return nullptr;
         }
     }
```

rbx: Not sure. But I don't want to touch those with this PR.

fairmq/shmem/Common.h
```diff
@@ -34,12 +35,21 @@ struct SharedMemoryError : std::runtime_error { using std::runtime_error::runtim
...
-    // boost::interprocess::iset_index>;
+    // boost::interprocess::null_index>;
+    boost::interprocess::iset_index>;
+
+struct RefCount
+{
+    std::atomic<uint16_t> count;
+};
+
+using MngRefCountPool = boost::interprocess::adaptive_pool<RefCount, boost::interprocess::managed_shared_memory::segment_manager, 1024, 2>;
```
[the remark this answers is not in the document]
rbx: What would be a good value for the upper bound?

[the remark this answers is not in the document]
rbx: I tried this:
```cpp
constexpr size_t maxNodesPerBlock = 4096;
constexpr size_t maxFreeBlocks = 2;

constexpr size_t nodesPerBlock() {
    return std::min(boost::interprocess::mapped_region::get_page_size() / sizeof(RefCount), maxNodesPerBlock);
}

constexpr auto numNodesPerBlock = nodesPerBlock();

using RefCountPool = boost::interprocess::adaptive_pool<RefCount, boost::interprocess::managed_shared_memory::segment_manager, nodesPerBlock, maxFreeBlocks>;
```
, but it fails the constexpr check:
`call to non-‘constexpr’ function ‘static std::size_t boost::interprocess::mapped_region::get_page_size()’ `.

Anyway, unmagicked the number a lil bit for now.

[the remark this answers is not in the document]
dennisklein: hm, from experience I know about 4kiB, 2MiB, and 1GiB as real-world page sizes used on 64bit x86 systems. I guess, 2MiB chunk would be fine still, but GiB is clearly too much ?! so, perhaps
```cpp
// size pool chunks at least as big as a memory page, but not larger than 2MiB
constexpr auto numRefCountsPerChunk = std::min(boost::interprocess::mapped_region::get_page_size(), 2 * 1024 * 1024) / sizeof(RefCount);
```
Additionally, one could make 2MiB the default upper bound, but make it a build-time configurable as well?

Just speaking out ideas loudly.

fairmq/shmem/Message.h
```diff
@@ -291,21 +295,28 @@ class Message final : public fair::mq::Message
...
+                fRegionPtr = fManager.GetRegionFromCache(otherMsg.fMeta.fRegionId);
+                if (fRegionPtr) {
+                    // create a reference couting object in for this message
+                    RefCount* rcPtr = fRegionPtr->MakeRefCount();
+                    // point the fShared in the unmanaged region message to the refCount object
+                    otherMsg.fMeta.fShared = fRegionPtr->HandleFromAddress(rcPtr);
+                    // point this message to the same content as the unmanaged region message
+                    fMeta = otherMsg.fMeta;
+                    // increment the refCount
+                    fRegionPtr->IncrementRefCount(rcPtr);
```
dennisklein, line 308:
```suggestion
                    rcPtr->Increment();
```

or am I overlooking any reason why this should be indirected through the `UnmanagedRegion` still?

fairmq/shmem/UnmanagedRegion.h
```diff
@@ -186,6 +187,23 @@ struct UnmanagedRegion
...
+
+    void RemoveRefCount(RefCount* refCount)
+    {
+        refCount->~RefCount();
+        fRefCountPool->deallocate(refCount, 1);
+    }
+
+    uint16_t GetRefCount(RefCount* refCount) { return refCount->Get(); }
+    uint16_t IncrementRefCount(RefCount* refCount) { return refCount->Increment(); }
+    uint16_t DecrementRefCount(RefCount* refCount) { return refCount->Decrement(); }
```
dennisklein, line 205: Are those still needed at all?

rbx: no

fairmq/shmem/UnmanagedRegion.h
```diff
@@ -245,12 +269,18 @@ struct UnmanagedRegion
...
+    std::unique_ptr<RefCountPool> fRefCountPool;
 
     std::thread fAcksReceiver;
     std::thread fAcksSender;
     RegionCallback fCallback;
     RegionBulkCallback fBulkCallback;
 
+    static std::string MakeSegmentName(const std::string& shmId, std::string_view segment, int regionIndex) {
+        return tools::ToString("fmq_", shmId, "_", segment, "_", regionIndex);
+    }
```
dennisklein, line 282: I realize now, you generate these segment names in many more places. I guess, I would then move this helper to `Common.h` and use it everywhere for the purpose of deduplicating the segment name logic to a single place.

rbx: Sure

dennisklein: ah, and note I sneaked in `std::string_view` for the second argument, not sure if we include the `<string_view>` header already.

fairmq/shmem/Message.h
```diff
@@ -266,8 +266,12 @@ class Message final : public fair::mq::Message
...
                 return 1;
             } else {
-                fManager.GetSegment(fMeta.fSegmentId);
-                return ShmHeader::RefCount(fManager.GetAddressFromHandle(fMeta.fShared, fMeta.fSegmentId));
+                fRegionPtr = fManager.GetRegionFromCache(fMeta.fRegionId);
+                if (fRegionPtr) {
+                    return fRegionPtr->GetRefCount(fRegionPtr->GetRefCountAddressFromHandle(fMeta.fShared));
+                } else {
+                    throw TransportError(tools::ToString("Cannot get unmanaged region with id ", fMeta.fRegionId));
+                }
```
dennisklein, line 274:
```suggestion
            if (fMeta.fShared < 0) { // UR msg is not yet shared
                return 1;
            }
            fRegionPtr = fManager.GetRegionFromCache(fMeta.fRegionId);
            if (!fRegionPtr) {
                throw TransportError(tools::ToString("Cannot get unmanaged region with id ", fMeta.fRegionId));
            }
            return fRegionPtr->GetRefCount(fRegionPtr->GetRefCountAddressFromHandle(fMeta.fShared));
```
`readability-else-after-return`

fairmq/shmem/Common.h
```diff
@@ -69,13 +176,24 @@ struct RegionInfo
         , fCreationFlags(flags)
         , fUserFlags(userFlags)
         , fSize(size)
+        , fRCSegmentSize(10000000)
+        , fDestroyed(false)
+    {}
```
dennisklein, line 181: Could be shortened with a delegating ctor (same with the one ctor above):

```cpp
    RegionInfo(const char* path, int flags, uint64_t userFlags, uint64_t size, const VoidAlloc& alloc)
        : RegionInfo(path, flags, userFlags, size, 10000000, alloc)
    {}
```

And since `10000000` is used twice, I suggest to deduplicate, e.g.
```cpp
    constexpr uint64_t DefaultRcSegmentSize = 10000000;    
```

fairmq/shmem/UnmanagedRegion.h
rbx: actually, I'm not sure what the impact is of replacing a simple constructor arguments with our ToString tool is, since the tool constructs a stringstream, among other things. For this contructor I don't care since it is called once. But some of the other usage (in the Monitor) is potentially called in a loop, so I would rather keep it minimal there.

fairmq/shmem/Common.h
dennisklein: alternative would be to use default member initializers:
```cpp
struct RegionInfo
{
    RegionInfo(const VoidAlloc& alloc)
        : fPath("", alloc)

    RegionInfo(const char* path, int flags, uint64_t userFlags, uint64_t size, const VoidAlloc& alloc)
        : fPath(path, alloc)
        , fCreationFlags(flags)
        , fUserFlags(userFlags)
        , fSize(size)
    {}

    RegionInfo(const char* path, int flags, uint64_t userFlags, uint64_t size, uint64_t rcSegmentSize, const VoidAlloc& alloc)
        : fPath(path, alloc)
        , fCreationFlags(flags)
        , fUserFlags(userFlags)
        , fSize(size)
        , fRCSegmentSize(rcSegmentSize)
    {}

    Str fPath;
    int fCreationFlags{0}; // = 0; should work as well
    uint64_t fUserFlags{0};
    uint64_t fSize{0};
    uint64_t fRCSegmentSize{10000000};
    bool fDestroyed{false};
};
```

fairmq/shmem/UnmanagedRegion.h
dennisklein: Or you could not use `tools::ToString` in the helper, and use the `+` operators just to be sure. My point would be to not duplicate the naming logic and the constant prefix etc. Easy to mess up/overlook in future refactorings etc.

dennisklein: Side note: In general, I would vote to adopt libfmt and modernize those types of string generation - I think we have a couple of places that could benefit from this. That's of course out of scope of this PR. The other day I had a look at [`std::print`](https://en.cppreference.com/w/cpp/io/print) (C++23) and realized it is not even implemented in GCC yet. So, this will be many years in the future still, so an adoption of libfmt seems worth.

fairmq/shmem/Common.h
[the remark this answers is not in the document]
dennisklein: > , but it fails the constexpr check:
> `call to non-‘constexpr’ function ‘static std::size_t boost::interprocess::mapped_region::get_page_size()’`

Didn't see this post, ya, we should detect the page size via CMake instead of looking it up at run-time.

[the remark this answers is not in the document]
dennisklein: hmm, this will not be enough. It could be different page sizes even for different regions. Also the `mapped_region::get_page_size()` is not the right way. As a consequence shmem page sizes will not be available at compile-time to look up.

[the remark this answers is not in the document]
rbx: I'm leaving it at 4096.

fairmq/shmem/Message.h
```diff
@@ -291,21 +292,28 @@ class Message final : public fair::mq::Message
...
+                    RefCount* rcPtr = fRegionPtr->MakeRefCount();
+                    // point the fShared in the unmanaged region message to the refCount object
+                    otherMsg.fMeta.fShared = fRegionPtr->HandleFromAddress(rcPtr);
+                    // point this message to the same content as the unmanaged region message
+                    fMeta = otherMsg.fMeta;
+                    // increment the refCount
+                    rcPtr->Increment();
+                } else {
+                    throw TransportError(tools::ToString("Cannot get unmanaged region with id ", fMeta.fRegionId));
+                }
```
dennisklein, line 308:
```suggestion
                fRegionPtr = fManager.GetRegionFromCache(otherMsg.fMeta.fRegionId);
                if (!fRegionPtr) {
                    throw TransportError(tools::ToString("Cannot get unmanaged region with id ", fMeta.fRegionId));
                }
                
                // create a reference couting object in for this message
                RefCount* rcPtr = fRegionPtr->MakeRefCount();
                // point the fShared in the unmanaged region message to the refCount object
                otherMsg.fMeta.fShared = fRegionPtr->HandleFromAddress(rcPtr);
                // point this message to the same content as the unmanaged region message
                fMeta = otherMsg.fMeta;
                // increment the refCount
                rcPtr->Increment();
```

On the other hand, if we always have this error handling anyways (which only depends on the `regionId`), we should move it to the getter, no?
```cpp
UnmanagedRegion& GetRegionFromCacheOrThrow(uint16_t id) {
    auto region = GetRegionFromCache(id); // or even directly throw from inside here?
    if (!fRegionPtr) {
       throw TransportError(tools::ToString("Cannot get unmanaged region with id ", fMeta.fRegionId));
    }
    return *region;
}             
```

fairmq/shmem/UnmanagedRegion.h
```diff
@@ -186,6 +187,19 @@ struct UnmanagedRegion
 
     bool RemoveOnDestruction() { return fRemoveOnDestruction; }
 
+    RefCount* MakeRefCount()
+    {
+        RefCount* refCount = fRefCountPool->allocate(1).get();
+        new (refCount) RefCount(1);
+        return refCount;
+    }
```
dennisklein, line 195:
```suggestion
    RefCount& MakeRefCount()
    {
        RefCount* refCount = fRefCountPool->allocate(1).get();
        new (refCount) RefCount(1);
        return *refCount;
    }
```

If we always return a valid `RefCount`, then let's not use raw pointer, which will technically force a caller to always check for `nullptr`.

fairmq/shmem/UnmanagedRegion.h
```diff
@@ -186,6 +187,19 @@ struct UnmanagedRegion
...
+
+    void RemoveRefCount(RefCount* refCount)
+    {
+        refCount->~RefCount();
+        fRefCountPool->deallocate(refCount, 1);
+    }
```
dennisklein, line 201:
```suggestion
    void RemoveRefCount(RefCount& refCount)
    {
        refCount.~RefCount();
        fRefCountPool->deallocate(&refCount, 1);
    }
```

or alternatively

```suggestion
    void RemoveRefCount(RefCount* refCount)
    {
        assert(refCount);
        refCount->~RefCount();
        fRefCountPool->deallocate(&refCount, 1);
    }
```

fairmq/shmem/Common.h
```diff
@@ -41,13 +44,126 @@ using RBTreeBestFitSegment = boost::interprocess::basic_managed_shared_memory<ch
...
+struct ShmHeader
+{
+    struct Hdr
+    {
+        uint16_t userOffset;
+        std::atomic<uint16_t> refCount;
+    };
+
+    static Hdr* HdrPtr(char* ptr)
+    {
```
dennisklein, line 95:
```suggestion
    static Hdr* HdrPtr(char* ptr)
    {
        assert(ptr);
```

Basically for all of the member functions that take raw pointer args and assume they are non-null, right?

fairmq/shmem/Message.h
rbx: I applied the suggestion. Didn't do the getter for now.

fairmq/shmem/UnmanagedRegion.h
rbx: Done

fairmq/shmem/Common.h
rbx: I don't wanna touch this part in this PR. It was just move from one place to another.

fairmq/shmem/UnmanagedRegion.h
rbx: Done

fairmq/shmem/Message.h
```diff
@@ -291,21 +292,27 @@ class Message final : public fair::mq::Message
...
-                ShmHeader::IncrementRefCount(fManager.GetAddressFromHandle(fMeta.fShared, fMeta.fSegmentId));
+                fRegionPtr = fManager.GetRegionFromCache(fMeta.fRegionId);
+                if (fRegionPtr) {
+                    fRegionPtr->GetRefCountAddressFromHandle(fMeta.fShared)->Increment();
+                } else {
+                    throw TransportError(tools::ToString("Cannot get unmanaged region with id ", fMeta.fRegionId));
+                }
             }
         }
     }
```
dennisklein, line 318: One more optimization by initializing the ref count with `2` and dedup `Copy` a bit for readability:

```diff
--- a/fairmq/shmem/Message.h
+++ b/fairmq/shmem/Message.h
@@ -276,45 +276,37 @@ class Message final : public fair::mq::Message
     void Copy(const fair::mq::Message& other) override
     {
         const Message& otherMsg = static_cast<const Message&>(other);
+        // if the other message is not initialized, close this one too and return
         if (otherMsg.fMeta.fHandle < 0) {
-            // if the other message is not initialized, close this one too and return
             CloseMessage();
             return;
         }

+        // if this msg is already initialized, close it first
         if (fMeta.fHandle >= 0) {
-            // if this msg is already initialized, close it first
             CloseMessage();
         }

-        if (otherMsg.fMeta.fManaged) { // managed segment
-            fMeta = otherMsg.fMeta;
-            fManager.GetSegment(fMeta.fSegmentId);
-            ShmHeader::IncrementRefCount(fManager.GetAddressFromHandle(fMeta.fHandle, fMeta.fSegmentId));
-        } else { // unmanaged region
+        // increment ref count
+        if (otherMsg.fMeta.fManaged) { // msg in managed segment
+            fManager.GetSegment(otherMsg.fMeta.fSegmentId);
+            ShmHeader::IncrementRefCount(fManager.GetAddressFromHandle(otherMsg.fMeta.fHandle, otherMsg.fMeta.fSegmentId));
+        } else { // msg in unmanaged region
+            fRegionPtr = fManager.GetRegionFromCache(otherMsg.fMeta.fRegionId);
+            if (!fRegionPtr) {
+                throw TransportError(tools::ToString("Cannot get unmanaged region with id ", fMeta.fRegionId));
+            }
+
             if (otherMsg.fMeta.fShared < 0) {
-                fRegionPtr = fManager.GetRegionFromCache(otherMsg.fMeta.fRegionId);
-                if (!fRegionPtr) {
-                    throw TransportError(tools::ToString("Cannot get unmanaged region with id ", fMeta.fRegionId));
-                }
-                // create a reference couting object in for this message
-                RefCount& rcPtr = fRegionPtr->MakeRefCount();
-                // point the fShared in the unmanaged region message to the refCount object
-                otherMsg.fMeta.fShared = fRegionPtr->HandleFromAddress(&rcPtr);
-                // point this message to the same content as the unmanaged region message
-                fMeta = otherMsg.fMeta;
-                // increment the refCount
-                rcPtr.Increment();
-            } else { // if the UR msg is already shared
-                fMeta = otherMsg.fMeta;
-                fRegionPtr = fManager.GetRegionFromCache(fMeta.fRegionId);
-                if (fRegionPtr) {
-                    fRegionPtr->GetRefCountAddressFromHandle(fMeta.fShared)->Increment();
-                } else {
-                    throw TransportError(tools::ToString("Cannot get unmanaged region with id ", fMeta.fRegionId));
-                }
+                // UR msg not yet shared, create the reference counting object with count 2
+                otherMsg.fMeta.fShared = fRegionPtr->HandleFromAddress(&(fRegionPtr->MakeRefCount(2)));
+            } else {
+                fRegionPtr->GetRefCountAddressFromHandle(otherMsg.fMeta.fShared)->Increment();
             }
         }
+
+        // copy meta data
+        fMeta = otherMsg.fMeta;
     }

     ~Message() override { CloseMessage(); }
diff --git a/fairmq/shmem/UnmanagedRegion.h b/fairmq/shmem/UnmanagedRegion.h
index 539a93822..a1b147f5c 100644
--- a/fairmq/shmem/UnmanagedRegion.h
+++ b/fairmq/shmem/UnmanagedRegion.h
@@ -187,10 +187,10 @@ struct UnmanagedRegion

     bool RemoveOnDestruction() { return fRemoveOnDestruction; }

-    RefCount& MakeRefCount()
+    RefCount& MakeRefCount(uint16_t initial_count = 1)
     {
         RefCount* refCount = fRefCountPool->allocate(1).get();
-        new (refCount) RefCount(1);
+        new (refCount) RefCount(initial_count);
         return *refCount;
     }

```

rbx: Nice, updated.

fairmq/shmem/UnmanagedRegion.h
```diff
@@ -245,12 +265,18 @@ struct UnmanagedRegion
...
+    std::unique_ptr<boost::interprocess::managed_shared_memory> fRefCountSegment;
+    std::unique_ptr<RefCountPool> fRefCountPool;
 
     std::thread fAcksReceiver;
     std::thread fAcksSender;
     RegionCallback fCallback;
     RegionBulkCallback fBulkCallback;
 
+    static std::string MakeSegmentName(const std::string& shmId, std::string_view segment, int regionIndex) {
+        return tools::ToString("fmq_", shmId, "_", segment, "_", regionIndex);
```
dennisklein, line 277: No longer used, or?

dennisklein (approved):

PR #485, opened 2023-10-18
Refactor shmem::Message to reduce its size in memory

rbx:
- Move the members of MetaHeader flat into shmem::Message and sort them by size.
- Remove alignment member. Accept explicit argument for the method where it is needed (SetUsedSize), optionally. By default try to deduce reasonable alignment.

This reduces the size of shmem::Message from 92 bytes to 80 bytes. Two more members not clearly visible when skimming through the class are the this pointer, since we use virtual methods, and the member of the parent class - `fTransport`.

### DIFF
--- a/fairmq/Message.h
+++ b/fairmq/Message.h
@@ -46,7 +46,7 @@ struct Message
     virtual void* GetData() const = 0;
     virtual size_t GetSize() const = 0;
 
-    virtual bool SetUsedSize(size_t size) = 0;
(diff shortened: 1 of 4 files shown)


rbx:
Want to do couple more tests before review, but let the CI run already.

dennisklein (approved):
Please put the explanation text from the PR why you flatten the MetaHeader member also into the first commit msg. Thx! Otherwise looks all fine to me.

PR #486, opened 2023-10-21
Drop unused variable

ktf:
Drop unused variable

The else clause at the end makes the postincrement impossible.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/FairRootGroup/FairMQ/pull/486).
* #487
* __->__ #486

### DIFF
--- a/fairmq/Device.cxx
+++ b/fairmq/Device.cxx
@@ -176,7 +176,6 @@ void Device::InitWrapper()
 
     // Fill the uninitialized channel containers
     for (auto& channel : GetChannels()) {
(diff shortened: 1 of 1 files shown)


rbx:
> Drop unused variable

agreed

> The else clause at the end makes the postincrement impossible.

it's possible if method is "bind" or "connect" or address contains "@+>"?

rbx (approved):

ktf:
> it's possible if method is "bind" or "connect" or address contains "@+>"?

Yes, of course...

PR #490, opened 2023-11-15
Ref count segment error handling

rbx:
- shm: Throw RefCountBadAlloc when insufficient space in the ref count segment
- shm: Bump the ref segment size 10x.
- Examples: use multipart in the region example


fairmq/shmem/Message.h
```diff
@@ -279,7 +279,11 @@ class Message final : public fair::mq::Message
             }
             if (otherMsg.fShared < 0) {
                 // UR msg not yet shared, create the reference counting object with count 2
-                otherMsg.fShared = fRegionPtr->HandleFromAddress(&(fRegionPtr->MakeRefCount(2)));
+                try {
+                    otherMsg.fShared = fRegionPtr->HandleFromAddress(&(fRegionPtr->MakeRefCount(2)));
+                } catch (boost::interprocess::bad_alloc& ba) {
+                    throw RefCountBadAlloc("Insufficient space in the reference count segment");
+                }
```
dennisklein, line 286: Is it just one segment or would it be useful to add some segment id to the message?

PR #500, opened 2024-01-12
Update MacOS CI entires

rbx:


rbx:
The CI jobs land as expected on the macs. The failure of some shmem tests on one of the macs is not understood, but is out of scope of this pr.

Issue #443, opened 2022-08-26
Add a simple switch to disable built-in control logic


dennisklein:
1. `--control none` (which would still allow to have `--catch-signals 1`)
2. `bool DeviceRunner::fDoNotLoadBuiltinControlPlugin` (no need to add a CLI option)

PR #503, opened 2024-02-19
Add 'no control' controller

rbx:
For #443.

Here I skip also the startup sequence. The plugin changes no states at all (device remains in Idle). Was that what we wanted to solve with this?

I've tested if the device still responds to signals with this and it does.

### DIFF
--- a/fairmq/plugins/control/Control.cxx
+++ b/fairmq/plugins/control/Control.cxx
@@ -65,13 +65,17 @@ Control::Control(const string& name, Plugin::Version version, const string& main
     });
 
     try {
-        TakeDeviceControl();
-
         auto control = GetProperty<string>("control");
+
+        if (control != "none") {
+            TakeDeviceControl();
+        }
 
         if (control == "static") {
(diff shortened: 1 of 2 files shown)


fairmq/plugins/control/Control.cxx
```diff
@@ -65,13 +65,18 @@ Control::Control(const string& name, Plugin::Version version, const string& main
...
+        if (control != "none") {
+            TakeDeviceControl();
+        }
+
         if (control == "static") {
             LOG(debug) << "Running builtin controller: static";
             fControllerThread = thread(&Control::StaticMode, this);
+        } else if (control == "none") {
+            LOG(debug) << "Running builtin controller: none";
+            fControllerThread = thread(&Control::NoControlMode, this);
```
dennisklein, line 79: Hm, let's not start a control thread at all?

rbx: Done!

dennisklein (approved):

PR #508, opened 2024-05-23
Use a SmallVector<T> as parts container

ktf:


ktf:
Will comment more soon.

PR #518, opened 2025-01-06
Out of line ProgOption::SetProperty for int and std::string

ktf:
The specializations are common enough to show up in O2 compilation profiles and they are not time critical (once per run at max).


ktf:
@rbx @dennisklein I have something similar for EventManager which also suffer from long compilation times. What do you think?

rbx:
Looks like it could be a nice optimization for the compilation speed for you. How much does it bring (just curious)?

dennisklein:
> For the record, there seems to be some issues with the testing infrastructure at GSI.

The Jenkinsfile in `master` branch is outdated so the CI failed. Safer to open against `dev` branch :)

ktf:
I opened it against whatever branch it gave me by default. I will reopen against dev.

Tests aside, would you be ok with such change? Can it be backported to the version which we use in ALICE?

ktf:
In limited tests, the whole FairMQ compilation which happens inside O2 (i.e. the stuff which is inlined in templates) accounts for 10% of the framework compilation time (roughly 70/700 seconds, including unit tests and test workflows). This particular one seems to save ~5s on my M1 mac.

ktf:
(to be clear, the single compilation is 100ms, the problem is that it's done in each single test / workflow)

ktf:
Profiled via -ftime-trace and ClangBuildAnalyzer.

dennisklein (approved):

PR #519, opened 2025-01-07
Reduce bloat due to statics

ktf:
Avoid disseminating every compile unit including Message.h with TransportNames and
TransportTypes and the associated unordered_map helper methods (e.g.
murmur_hash).


dennisklein (approved):
Looks fine to me. Thx and a happy new year :)

PR #532, opened 2025-06-10
De-boostify: use std::pmr from C++17

ktf:


ktf:
Actually it looks like it's in C++17 already, maybe we can even remove the ifdef, then?

dennisklein:
* :+1: to de-boostify
* I wouldn't mind, if we bumped required C++ to C++20.
* If you can confirm this works with the Alice-relevant GCC and (Apple)Clang versions, let's remove the #ifdef.

ktf:
I am working on making O2 able to digest both. Apparently there are some subtle changes which require minor adaptations, see: https://github.com/AliceO2Group/AliceO2/pull/14384 and https://github.com/AliceO2Group/AliceO2/pull/14379. I am now checking everything is in order and let you know.

ktf:
https://github.com/alisw/alidist/pull/5922 is the last remaining bit. If this passes tests, I would claim we (ALICE) can use both `std::pmr` and `boost::container::pmr` independently. Once that's validated, I would kindly ask you for an official tag with this.

ktf:
Actually, it looks like trying with a private commit confuses CMake. Can you please tag this? Otherwise I get:

```
## sw/BUILD/Control-OCCPlugin-latest/log
--
CMake Error at CMakeLists.txt:154 (find_package):
  Could not find a configuration file for package "FairMQ" that is compatible
  with requested version "1.4.2".

  The following configuration files were considered but not accepted:

    /sw/slc9_x86-64/FairMQ/v1.10.0-local1/lib/cmake/FairMQ-1.2.2/FairMQConfig.cmake, version: 1.2.2
```

when I use a hash.

PR #548, opened 2026-06-01
ci: cache gcc as a buildcache node

dennisklein:
Follow-up to #547. Makes CI builds fast by caching the gcc compiler.

## Problem
gcc was registered as an external via `spack compiler find`, which excludes it from `buildcache push` — so gcc was never cached and every CI job rebuilt it from source (~1 h). The earlier committed-lockfile attempt pinned gcc as a host-path external, which isn't portable across runners and broke CI.

## Change
- push the freshly-built gcc node in `setup-deps` **before** `spack compiler find` (which marks it external), gated behind a `push-gcc` input used only by the buildcache workflow
- drop the committed-lockfile approach: remove `test/ci/locks`, the lockfile install path in `setup-deps`, and the lockfile export in the buildcache workflow
- drop the ignored `ref` input from `setup-spack` (v3 renamed it to `spack_ref`)

## Validation
Probed on a fork: a fresh runner pulls gcc@15 from the buildcache in ~50 s instead of building it ~1 h:
```
[ ] uythfib gcc@15.2.0 fetching from build cache (0s)
[+] uythfib gcc@15.2.0 .../gcc-15.2.0-uythfib... (20s)
real 0m51.033s
```

## Rollout
After merge, the push to `dev` triggers the buildcache workflow (it touches `setup-deps/**` + `buildcache.yml`), which builds and **pushes the gcc nodes** for each matrix entry. Subsequent CI runs then pull gcc instead of building it. (CI on this PR still builds gcc from source until that first buildcache run populates the cache.)


dennisklein:
Holding as draft. CI investigation shows the committed lockfiles pin gcc as a host-path `external` (from `spack compiler find`), which doesn't transfer between runners — CI fails at CMake configure with `CMAKE_CXX_COMPILER not set`.

Root cause is broader: gcc is compiled from source (~58 min) on every buildcache run and never pushed (externals are excluded from `buildcache push`), so it's not cacheable and the lockfiles can't be portable. Investigating how to make the spack-built gcc a cacheable node before reworking this PR.

PR #555, opened 2026-06-10
feat(shmem): expose side-channel metadata API for unsent messages

rbx:
Add two public entry points needed by the ALICE use case where shmem messages are allocated via a transport but never sent - their metadata is instead serialised into Arrow tables and delivered over a separate channel, allowing consumer devices to resolve the payload pointer without taking ownership.

`shmem::Message::GetMeta()` returns the MetaHeader of the message.

`shmem::GetDataAddressFromHandle(TransportFactory&, const MetaHeader&)` is a free function declared in Common.h and defined in Manager.cxx. Keeping it out of the TransportFactory class body means callers only need to include Common.h (available transitively via Message.h) and do not drag in Socket.h or zmq.h.

The implementation handles both managed segments and unmanaged regions, and throws `SharedMemoryError` with a typed message on a bad segment or region id.

Lifetime of the returned pointer is the caller's responsibility; the cache device is expected to hold the messages alive.

### DIFF
--- a/fairmq/shmem/Common.h
+++ b/fairmq/shmem/Common.h
@@ -339,4 +339,13 @@ struct SegmentBufferShrink
 
 } // namespace fair::mq::shmem
 
+namespace fair::mq { class TransportFactory; }
+
+namespace fair::mq::shmem {
+// Resolve a MetaHeader (received over a side channel) to the local data pointer.
+// The caller is responsible for ensuring the backing buffer remains alive for the
+// duration of access; FairMQ provides no refcount protection for this path.
+char* GetDataAddressFromHandle(fair::mq::TransportFactory& factory, const MetaHeader& meta);
+} // namespace fair::mq::shmem
+
 #endif /* FAIR_MQ_SHMEM_COMMON_H_ */
--- a/fairmq/shmem/Manager.cxx
+++ b/fairmq/shmem/Manager.cxx
@@ -7,6 +7,7 @@
  ********************************************************************************/
 
 #include "Manager.h"
+#include "TransportFactory.h"
 
 // Needed to compile-firewall the <boost/process/async.hpp> header because it
 // interferes with the <asio/buffer.hpp> header. So, let's factor
@@ -49,6 +50,14 @@ bool Manager::SpawnShmMonitor(const std::string& id)
         exe, "-x", "-m", "--shmid", id, "-d", "-t", "2000", (verbose ? "--verbose" : ""), env);
 
     return true;
+}
+
+char* GetDataAddressFromHandle(fair::mq::TransportFactory& factory, const MetaHeader& meta)
+{
+    if (factory.GetType() != fair::mq::Transport::SHM) {
+        throw SharedMemoryError("GetDataAddressFromHandle called on a non-shmem transport");
+    }
+    return static_cast<TransportFactory&>(factory).GetDataAddressFromHandle(meta);
 }
 
 }   // namespace fair::mq::shmem
--- a/fairmq/shmem/Manager.h
+++ b/fairmq/shmem/Manager.h
@@ -776,6 +776,30 @@ class Manager
 
     auto GetMetadataMsgSize() const noexcept { return fMetadataMsgSize; }
 
+    // Resolve a MetaHeader (received over a side channel) to the local data pointer.
+    // The caller is responsible for ensuring the backing buffer remains alive for the
+    // duration of access; FairMQ provides no refcount protection for this path.
+    char* GetDataAddressFromHandle(const MetaHeader& meta)
+    {
+        if (meta.fManaged) {
+            if (meta.fSize == 0) {
+                return nullptr;
+            }
+            GetSegment(meta.fSegmentId);
+            auto it = fSegments.find(meta.fSegmentId);
+            if (it == fSegments.end()) {
+                throw SharedMemoryError(tools::ToString("GetDataAddressFromHandle: cannot open segment with id ", meta.fSegmentId));
+            }
+            return ShmHeader::UserPtr(GetAddressFromHandle(meta.fHandle, meta.fSegmentId));
+        } else {
+            UnmanagedRegion* region = GetRegionFromCache(meta.fRegionId);
+            if (!region) {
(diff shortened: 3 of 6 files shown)


dennisklein (commented):
> [!NOTE]
> This review was generated with Claude.

### Overview

Adds two public entry points so ALICE-style consumers can resolve shmem payload pointers from metadata delivered out-of-band (e.g. serialized into Arrow tables) instead of via FairMQ channels:

- `shmem::Message::GetMeta()` — returns the message's `MetaHeader` by value.
- `shmem::GetDataAddressFromHandle(TransportFactory&, const MetaHeader&)` — free function declared in `Common.h`, defined in `Manager.cxx`, dispatching through `shmem::TransportFactory` → `Manager`.

+158/−0 across 6 files, three new tests. No API breaks, purely additive.

### Correctness — verified against the codebase, looks right

- **`GetMeta()` aggregate init order matches `MetaHeader`** exactly (`fSize, fHint, fHandle, fShared, fRegionId, fSegmentId, fManaged`). ✔
- **`Manager::GetDataAddressFromHandle()` faithfully mirrors `Message::GetData()`** for all three paths: zero-size managed → `nullptr`, managed → `ShmHeader::UserPtr(GetAddressFromHandle(...))`, unmanaged → `region->GetData() + fHandle`. ✔
- It's actually **safer than `GetData()` on failure**: `GetSegment()` swallows open failures with just a `LOG(error)`, after which `GetData()` would die on an uncaught `std::out_of_range` from `fSegments.at()`. The new code re-checks the map and throws a typed `SharedMemoryError` instead. Good.
- The `static_cast` downcast in the free function is guarded by the `GetType() != Transport::SHM` check, and `shmem::TransportFactory` is `final`. ✔
- Returning a raw pointer rather than constructing a `Message` from the header is the right call — a real `shmem::Message` would deallocate/deref-count the buffer on destruction, which is exactly what this no-ownership path must avoid.
- The `#include "TransportFactory.h"` in Manager.cxx doesn't break the boost/process compile-firewall — that comment is about keeping boost/process out of *headers*; this include only flows zmq.h into a .cxx that's already heavy.

### Issues

**1. CI is red because of an unused parameter (must fix).** `SideChannelUnmanaged(const string& address)` never uses `address` — unlike the neighboring tests, it never binds/connects a channel. The `static-analysis` job fails on exactly this (`_message.cxx:497: -Wunused-parameter` / `misc-unused-parameters`). Drop the parameter and the call-site argument; all other CI jobs (gcc 12–15, asan/ubsan, tsan, boost 1.87) are green.

**2. Negative caching makes the unmanaged error path permanently sticky — worth documenting.** `GetRegionFromCache()` caches `nullptr` results in the thread-local cache, and the generation counter `fRegionsGen` only bumps on *local* region create/remove. So if a consumer resolves a MetaHeader before the producer's region is discoverable, the `SharedMemoryError` will keep being thrown on that thread forever — retrying never heals. This is pre-existing behavior shared with `GetData()`, but this PR is the first API explicitly designed for cross-process resolution where arrival order isn't guaranteed, so the doc comment should warn that the call is not retryable on region-not-found (or the cache should skip caching `nullptr`).

**3. Tests never exercise the actual cross-factory use case.** Both round-trip tests resolve metadata through the *same* factory that allocated the message, so the segment/region is already in the local maps — the `open_only` attach paths (`GetSegment`'s lazy open, `GetRegion`'s remote lookup) are never hit through the new API, yet that's the entire production scenario. Suggest creating a second factory with the same session id and resolving there.

### Minor / nits

- **Missing error test for the unmanaged branch**: a `fManaged = false`, bogus `fRegionId` case (`EXPECT_THROW(..., SharedMemoryError)`) would complete the error matrix in `SideChannelErrors()`.
- **Name**: `GetDataAddressFromHandle` takes a `MetaHeader`, not a handle (unlike `Manager::GetAddressFromHandle`, which does). `GetDataAddressFromMeta` would be more accurate — worth considering before the API is set in stone.
- `GetMeta()`'s positional aggregate init silently misassigns if `MetaHeader` fields are ever reordered (the size_t/handle_t/uint16_t pairs would still compile). Low risk since `MetaHeader` is the wire format, but a brief comment on the dependency wouldn't hurt.
- The bad-segment-id test will emit a `LOG(error)` from `GetSegment` before the throw — cosmetic noise in test output, fine to leave.

### Verdict

The implementation is correct and the design rationale (free function in `Common.h` to avoid dragging zmq.h into consumers; raw pointer to avoid ownership) is sound. **Needs work before merge**: fix the unused `address` parameter (it's the only thing failing CI), and ideally add a second-factory test plus a doc note on the non-retryable region-not-found behavior.

dennisklein (approved):

aalkin:
Thank you for preparing this implementation, indeed, this is cleaner than my attempt in #551. For my own curiosity, what is the reason for the implementations of TransportFactory and others to be header-only?

dennisklein:
> For my own curiosity, what is the reason for the implementations of TransportFactory and others to be header-only?

Originally, transport implementations [used to be private](https://github.com/FairRootGroup/FairMQ/commit/24e7a5b8d01aa30c51820f67c13a2cb7bc1ac7da). But over time more and more of the shmem details have been exposed instead of changing/redesigning the abstract interfaces. The changes to header-only (https://github.com/FairRootGroup/FairMQ/commit/ccbf0be572ba22882065f1878353fa1e54aad4ae, https://github.com/FairRootGroup/FairMQ/commit/df574c64665fd601084f3c2ecf926b7835d357c4) were performed when they were still private, unfortunately, the commit msgs do not explain it. Perhaps @rbx remembers. I guess just organizationally, fewer files.

rbx:
Generally, having everything available in the headers should give the compiler more room and information for optimization. That was the primary motivation back then, although it may be becoming less relevant today with various LTO techniques. Naturally, at some point the compile time will climb so it becomes a trade-off between optimization potential and compile time. In principle I am not against changing this if there is a good reason to do so. But that should go in a separate PR to keep it focused.

rbx:
@aalkin do you need a tag with this feature?

aalkin:
@rbx yes please, we would like to integrate it as soon as possible.

